PR #98, opened 2021-08-20
[keepercore] Allow graph merge updates

aquamatthias:
This PR implements a graph merge method and replaces the subgraph update method.

A merge graph is a self contained graph under a node which is marked with merge=true.
Such nodes are merged with the merge node in the database.
Example:
```
A -> B -> C(merge=true) -> E -> E1 -> E2
                        -> F -> F1
       -> D(merge=true) -> G -> G1 -> G2 -> G3 -> G4
```

Node `C` and `D` are marked as merge node: those nodes do not get replaced but merged.
3 updates are considered for node: `E`, `F` and `G` which will replace existing nodes with the same id.

Assume the existing graph in the database looks like this:
```
A -> B -> C -> Z -> Z1 -> Z2
            -> F -> F1
       -> D -> G -> G13
       -> D -> H -> H1

```

After the merge, the final graph would be this:
```
A -> B -> C(merge=true) -> E -> E1 -> E2
                        -> F -> F1
                        -> Z -> Z1 -> Z2
       -> D(merge=true) -> G -> G1 -> G2 -> G3 -> G4
                        -> H -> H1
```

This update mechanism is more powerful than the update_subgraph mechanism and is replacing it.


keepercore/core/model/graph_access.py
```diff
@@ -173,3 +184,53 @@ def root_id(graph: DiGraph) -> str:
         roots: List[str] = [n for n, d in graph.in_degree if d == 0]
         assert len(roots) == 1, f"Given subgraph has more than one root: {roots}"
         return roots[0]
+
+    @staticmethod
+    def merge_graphs(
```
aquamatthias, line 189: add documentation! there is a lot going on here.

PR #178, opened 2021-09-21
[keepercore] tag command needs to resolve cloud/account/region/zone and update the resulting node

### LINKED ISSUE #153: [keepercore] tag command needs to resolve cloud, account, region, zone
## Why
The tagger needs additional information for cloud, account, region and zone.
This information should be provided without further user interaction.

## What
Incoming elements need to restore load and merge all references.

## AC
`match | tag` command can be issued without any additional `merge_ancestors`

aquamatthias:
Fixes #153


lloesche (approved):
Works great!

PR #194, opened 2021-09-23
[keepercore] Fix null<any is true

### LINKED ISSUE #193: [keepercore] comparison with non existing values yields true.
comparison queries should yield `x!=null and x<y`

aquamatthias:
Fixes #193


lloesche (approved):
lgtm 🚢

PR #196, opened 2021-09-23
[keepercore] Add a text/plain output format

aquamatthias:


lloesche (approved):
json as default... 🙄 😂 🚢

PR #264, opened 2021-09-29
[docs] optimization for documentation autobuild

neilinger:
adjusted github actions to only run when needed


aquamatthias (approved):
Nice

PR #279, opened 2021-09-30
WIP - Pin all dependency versions and remove obsolete dependencies

lloesche:
- Pins all versions
- removes Pympler dependency
- removes defaultlist dependency


cklib/cklib/utils.py
```diff
@@ -633,3 +630,20 @@ def type_str(o):
     if module == "builtins":
         return cls.__qualname__
     return module + "." + cls.__qualname__
+
+
+class defaultlist(list):
```
aquamatthias, line 635: intended change?

aquamatthias (dismissed):
thanks!

cklib/cklib/utils.py
lloesche: Yes, I'm removing an external dependency to a module called `defaultlist` and replace it with a single class of identical function in `cklib.utils`.

PR #306, opened 2021-10-05
Blast Radius UI and cleanup

RPicster:
- A new blast radius ui was added. You will see it when clicking "mark for cleanup" on a node.
- The bottom command line was made smaller
- 'connections' were renamed to 'edges' in all scripts
- The graph view was refactored into a separate class, this will make it easier to use it in different ui elements (such as blast radius,  the main graph or a dashboard element)
- A problem with copying and duplicating graph data was fixed by making a true deep copy of the graph data.


lloesche (approved):
lgtm 🚢

PR #376, opened 2021-10-20
[ckworker/cklib] Optimize usage of deferred connections

lloesche:
Deferred connections are graph edges that could not be resolved at collect time.

This PR:
- adds an arg `--ignore-deferred-connections` which turns off the resolving of such during the graph sanitise.
- tries to resolve deferred connections early

Background:
Cloudkeeper collects accounts and regions in parallel, building a subgraph per region. This allows it to scale well with many CPU cores. However some edges go across regions (e.g. IAM instance profile to an instance in another region). For those deferred_connections were added to the graph. When a connection between two resources can not be made in the local subgraph, that connection is stored and later searched when all subgraphs have been merged into the global graph.

However this also means that worst case the final search happens single threaded on the global graph with O(n^2) complexity. In praxis this isn't the case since very few resources create deferred connections, but even a few thousand can slow down collect performance considerably.

### DIFF
--- a/cklib/cklib/graph/__init__.py
+++ b/cklib/cklib/graph/__init__.py
@@ -108,6 +108,7 @@ class Graph(networkx.DiGraph):
             self.add_edge(self.root, graph.root)
         else:
             log.warning("Merging graphs with no valid roots")
+        self.resolve_deferred_connections()
 
     def add_resource(self, parent, node_for_adding, **attr):
         """Add a resource node to the graph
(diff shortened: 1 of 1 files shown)


lloesche:
This PR does this actually. The ignore is just in addition to the optimization.
The `self.resolve_deferred_connections()` in the graph `merge()` method does exactly that. It tries to find deferred connections early in the account collection thread, before handing the cloud graph to the worker thread. So best case all the connections could already be resolved on the account level.

PR #425, opened 2021-11-16
[ckcore] Introduce dump command and use list output by default

aquamatthias:


lloesche (commented):
Works great! Thank you! Before we merge we should update the documentation to reflect this very substantial change in cksh usage.

lloesche:
Actually this seems to be borked now:
```
> match is(resource) | count

>
```

PR #431, opened 2021-11-24
[ckcore] Add a query template engine

aquamatthias:
This change allows for creating query templates and storing them in a query template library.
A query template is a named query with placeholders.
The placeholders can be replaced with specific values and create an executable query.

Templates can be tested and maintained via the `templates` CLI command:

```
$> templates test kind=volume is({{kind}})
is(volume)
$> templates add filter_kind is({{kind}})
Template filter_kind added to the query library.
is({{kind}})
> templates
filter_kind: is({{kind}})
$> templates filter_kind
is({{kind}})
$> templates delete filter_kind
Template filter_kind deleted from the query library.
```

Once a template is stored, it can be used in any query like this:
```
$> query expand(filter_kind, kind=volume) 
```

It is possible to expand and refine the query
```
$> query expand(filter_kind, kind=volume) and reported.name=~test limit 3
```

It is possible to use one or more expands in the same query.

```
$> query expand(filter_kind, kind=volume) <-- expand(filter_kind, kind=volume_type) limit 3
```


lloesche:
Works fine for me. We should decide whether we want to use plural or singular form for commands. `jobs` vs. `template`.
The doc above says `template test` but in my testing it was still `template expand`.

PR #432, opened 2021-11-24
[ckcore] is() allows for list of kinds

aquamatthias:
`is(volume)` matches all resources of type volume

`is([volume, instance])` matches all resources of type volume or of type instance


lloesche (approved):
Works great, ty!

PR #447, opened 2021-12-01
fix build k8s ci

yuval-k:


yuval-k:
the fix was to just use the latest kind version :)

PR #453, opened 2021-12-03
[ckcore] Add COOP/COEP headers

lloesche:
These are required for the UI to be able to use [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer).


aquamatthias (approved):
LGTM - thanks.

PR #461, opened 2021-12-07
[ckcore] Enable TLS secured endpoint

aquamatthias:


lloesche:
nit: should it be `tls` instead of `ssl`, since we (the world) are not using ssl anymore?

PR #466, opened 2021-12-08
[ckui] Networking fixes for web-ui

RPicster:
As chunks can contain multiple graph elements (compared to the desktop client where 1 chunk = 1 element), some adjustments had to be made.

### DIFF
--- a/ckui/src/ui/3delements/scripts/element_graphview_3D.gd
+++ b/ckui/src/ui/3delements/scripts/element_graphview_3D.gd
@@ -37,6 +37,7 @@ var root_node : Object = null
 var is_removed := false
 var is_active := true
 var update_visuals := false
+var size_by_descendants := false
 
 var node_group : Spatial = null
 var line_group : Spatial = null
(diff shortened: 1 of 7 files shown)


ckui/src/ui/3delements/scripts/element_graphview_3D.gd
```diff
@@ -122,22 +123,23 @@ func end_streaming():
...
-		
+	if size_by_descendants:
+		var _total_descendants:= 0.0
+		var descendants_values := []
 		for node in graph_data.nodes.values():
-			if "metadata" in node.data and "descendant_count" in node.data.metadata:
-				node.scene.descendant_scale = node.data.metadata.descendant_count / largest_descendant_value
+			if ("metadata" in node.data
+			and "descendant_count" in node.data.metadata
+			and (node.kind == "aws_account" or node.kind == "gcp_project")):
```
aquamatthias, line 132: we probably need something more generic..

aquamatthias (approved):
lgtm - thanks!

ckui/src/ui/3delements/scripts/element_graphview_3D.gd
RPicster: That whole block is inactive at the moment so it won't cause any issues for now.
But I agree that a more generic approach would be better.

This should also be solvable by using `get_children_count()` with the new structure.

PR #480, opened 2021-12-14
[ckcore] set host to 0.0.0.0 inside docker by default

### LINKED ISSUE #476: [docker] bind to 0.0.0.0 by default when running inside docker

aquamatthias:
Fixes #476


lloesche:
We should do this within the `docker/defaults` file instead of inside `ckcore`.

lloesche (changes requested):
Exporting `CKCORE_HOST=0.0.0.0` via `docker/defaults` and inside the `bootstrap` script should be enough to achieve this. No change in `ckcore` itself needed.

ckcore/core/__main__.py
```diff
@@ -109,7 +109,7 @@ async def on_start() -> None:
                 "created_at": to_json(system_data.created_at),
                 "system": platform.system(),
                 "platform": platform.platform(),
-                "inside_docker": in_docker,
+                "inside_docker": inside_docker,
```
lloesche, line 112: What is this being used for and how does it relate to binding on any interface?

lloesche (changes requested):

ckcore/core/__main__.py
aquamatthias: somehow unrelated - the variable existed. I just renamed it. Does not change any logic.

PR #482, opened 2021-12-15
[cklib] x509 functions for bootstrapping a CA and issuing certs

lloesche:
This PR adds the ability to bootstrap a CA, generate CSRs and sign them and write/read them to/from disk.

### DIFF
--- a/.github/workflows/check_pr_cklib.yml
+++ b/.github/workflows/check_pr_cklib.yml
@@ -38,3 +38,9 @@ jobs:
       - name: Run tests
         working-directory: ./cklib
         run: tox
+
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@v2
+        with:
+          name: code-coverage-report
+          path: cklib/htmlcov/
--- a/.github/workflows/check_pr_cklib.yml
+++ b/.github/workflows/check_pr_cklib.yml
@@ -42,5 +42,5 @@ jobs:
       - name: Archive code coverage results
         uses: actions/upload-artifact@v2
         with:
-          name: code-coverage-report
+          name: cklib-code-coverage-report
           path: cklib/htmlcov/
--- a/cklib/cklib/x509.py
+++ b/cklib/cklib/x509.py
@@ -27,11 +27,12 @@ def gen_rsa_key(key_size: int = 2048) -> RSAPrivateKey:
 
 def bootstrap_ca(
     days_valid: int = 3650,
-    common_name: str = "Cloudkeeper Certificate Authority",
+    common_name: str = "Cloudkeeper Root CA",
     organization_name: str = "Some Engineering Inc.",
     locality_name: str = "San Francisco",
     state_or_province_name: str = "California",
     country_name: str = "US",
+    path_length: int = 2,
 ) -> Tuple[RSAPrivateKey, Certificate]:
     ca_key = gen_rsa_key()
     subject = issuer = x509.Name(
@@ -53,7 +54,9 @@ def bootstrap_ca(
         .not_valid_after(
             datetime.utcnow().replace(tzinfo=timezone.utc) + timedelta(days=days_valid)
         )
-        .add_extension(x509.BasicConstraints(ca=True, path_length=None), critical=True)
+        .add_extension(
+            x509.BasicConstraints(ca=True, path_length=path_length), critical=True
+        )
         .add_extension(
             x509.KeyUsage(
                 digital_signature=False,
--- a/cklib/cklib/x509.py
+++ b/cklib/cklib/x509.py
@@ -1,0 +1,240 @@
+import os
+from ipaddress import (
+    ip_address,
+    ip_network,
+    IPv4Address,
+    IPv6Address,
+    IPv4Network,
+    IPv6Network,
+)
+from datetime import datetime, timedelta, timezone
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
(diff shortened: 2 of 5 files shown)


cklib/cklib/x509.py
```diff
@@ -0,0 +1,228 @@
...
+    csr_build = x509.CertificateSigningRequestBuilder().subject_name(
+        x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, common_name)])
+    )
+    if len(san_dns_names) + len(san_ip_addresses) > 0:
+        csr_build = csr_build.add_extension(
+            x509.SubjectAlternativeName(
+                [x509.DNSName(n) for n in san_dns_names]
+                + [x509.IPAddress(make_ip(i)) for i in san_ip_addresses]
+            ),
+            critical=False,
```
aquamatthias, line 95: why false?

cklib/cklib/x509.py
```diff
@@ -0,0 +1,228 @@
...
+        if not isinstance(extension.value, x509.SubjectAlternativeName):
+            continue
+        crt_build = crt_build.add_extension(
+            extension.value, critical=extension.critical
+        )
+    return crt_build.sign(ca_key, hashes.SHA256(), default_backend())
+
+
+def write_csr_to_file(csr: CertificateSigningRequest, csr_path: str) -> None:
+    return write_cert_to_file(csr, csr_path)
```
aquamatthias, line 152: a signing request is a certificate?

.github/workflows/check_pr_cklib.yml
```diff
@@ -38,3 +38,9 @@ jobs:
       - name: Run tests
         working-directory: ./cklib
         run: tox
+
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@v2
+        with:
+          name: cklib-code-coverage-report
+          path: cklib/htmlcov/
```
aquamatthias, line 46: what do we do with those reports?

aquamatthias (approved):
LGTM - I asked for a couple of comments for people that are not deeply into certificate handling (like me).

cklib/cklib/x509.py
```diff
@@ -0,0 +1,228 @@
...
+            x509.NameAttribute(NameOID.COMMON_NAME, common_name),
+        ]
+    )
+    ca_cert = (
+        x509.CertificateBuilder()
+        .subject_name(subject)
+        .issuer_name(issuer)
+        .public_key(ca_key.public_key())
+        .serial_number(x509.random_serial_number())
+        .not_valid_before(datetime.utcnow().replace(tzinfo=timezone.utc))
```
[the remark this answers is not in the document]
lloesche: No, `now()` gives your your local timezone based time, `utcnow()` gives you UTC. The replace of `tzinfo` just adds the TZ information to the `datetime` object. Without it you'll have a UTC datetime but the TZ won't be stored with the object.

cklib/cklib/x509.py
lloesche: Yes, it is a certificate of a special type signed with the private key of the cert requester. The resulting certificate is then signed with the private key of the CA but issued for the public key part of the private key that the request was signed with.

.github/workflows/check_pr_cklib.yml
lloesche: I used them to figure out the missing 2% to get 100% coverage. 😄

cklib/cklib/x509.py
lloesche: Because those are optional extensions and generally not critical if unsupported. As opposed to the base extensions above.

cklib/cklib/x509.py
```diff
@@ -0,0 +1,228 @@
...
+            x509.KeyUsage(
+                digital_signature=False,
+                key_encipherment=False,
+                key_cert_sign=True,
+                key_agreement=False,
+                content_commitment=False,
+                data_encipherment=False,
+                crl_sign=True,
+                encipher_only=False,
+                decipher_only=False,
```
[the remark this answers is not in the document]
lloesche: These are x509 capabilities universal. They define what the certificate is allowed and not allowed to do. Since this is the CA cert it's only allowed to sign other certs as well as the certificate revocation list. Further down where we sign the CSR you'll see that those certs are only allowed to encrypt traffic but not sign other certificates. If we wanted to have E.g. a root CA and then intermediate CAs we could issue certs with the capability to sign others and create a longer cert path. That way we'd have a train of trust.

cklib/cklib/x509.py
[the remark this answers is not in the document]
aquamatthias: mmh - I do not see the difference.
```
$ date
Wed 15 Dec 2021 12:31:26 PM CET
$ python
>>> from datetime import datetime, timezone
>>> datetime.now(timezone.utc)
datetime.datetime(2021, 12, 15, 11, 31, 37, 668133, tzinfo=datetime.timezone.utc)
>>> datetime.utcnow().replace(tzinfo=timezone.utc)
datetime.datetime(2021, 12, 15, 11, 31, 41, 260257, tzinfo=datetime.timezone.utc)
```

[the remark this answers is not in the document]
lloesche: Oh I overlooked that you added the timezone as an arg to `now()`. What I meant was this behaviour:

```
In [2]: datetime.now()
Out[2]: datetime.datetime(2021, 12, 15, 12, 36, 28, 902232)

In [3]: datetime.utcnow()
Out[3]: datetime.datetime(2021, 12, 15, 11, 36, 58, 578776)
```
Where the resulting `utcnow` generated datetime object doesn't have the UTC timezone info attached.

My bad, your change would result in the same output!

cklib/cklib/x509.py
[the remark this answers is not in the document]
aquamatthias: Sorry - I mean it would be great to have that comment in the code.

PR #483, opened 2021-12-15
[ckworker,ckmetrics,cksh,plugins] store pytest coverage reports as artifacts

lloesche:


aquamatthias (approved):
Fine with me. Maybe we should use a service like codecov.io which would give us better visibility.

PR #487, opened 2021-12-17
[ckui] Fix ancestor paths

aquamatthias:


RPicster:
Looks good! 👏
We also need to add the new ancestor structure to the example data file.

RPicster (approved):

PR #513, opened 2022-01-05
[ckcore] Implement http[s] command to integrate via web hooks

### LINKED ISSUE #149: [keepercore] Add http command to support web hooks
## Why
While it is possible to attach to the event bus, or to query the state of the system, it might be the simplest solution to also allow for web hooks to be triggered.
In order to support this scenario, a `http` command should be implemented, which will perform the request on execution.
This way it would be possible to trigger web hooks based on events (or any other trigger) in the system.

## What
- Add an http command that allows to specify: method, url
- Every element in the in_stream will trigger a http request
- When the method is POST: send the element as body
- Allow the definition of parallel requests per command

## Open Questions
- How to handle auth?

## Acceptance Criteria
TBD

aquamatthias:
TODO:
- [x] better testing
- [x] ideally https can be an alias to http - but then the command needs to know how it was called

This PR introduces the http[s] command that allows to call out to other services via the CLI.
This is useful in cases there is already automation in place, that can be triggered from cloudkeeper.

Use cases: 
- rules can be codified and trigger an endpoint, if a rule is violated
- custom resource handling service needs to be called in defined situations 
- information should be gathered after a specific event or action
- integrate with other services

Usage example:
```
$> query is(volume) and reported.volume_encrypted==false |  chunk 50 |  https violations.my.company.org/handle_unencrypted type==volume
```
Find all unencrypted volumes, chunk the resulting list by 50 elements and then send the chunks to this endpoint `https://violations.my.company.org/handle_unencrypted?type=volume"
The request body contains a json array with a list of unencrypted volumes.

Fixes #149


lloesche (approved):
LGTM 🚢

PR #541, opened 2022-01-14
[resotocore] query and CLI react on configurable section name

aquamatthias:


lloesche:
Seems to work for metadata but not ancestors?
```
> query instance_cores > 8 | tail -1 | format {name} {instance_cores} {/ancenstors.cloud.reported.name} {/metadata.cleaned}
teamcity-production-nov-19-2020 16 null False
```

lloesche (approved):
works fine for me

PR #551, opened 2022-01-18
[docker] Split of Docker images

lloesche:
This splits the build of the Docker image into multiple separate images.
Also adds support for building on multiple architectures (amd64 and arm64).
ARM images are only build for tagged versions as the QEMU based build on Github's x86 machines takes about 4h to complete vs. 16min for the standard image.


.github/workflows/split-docker-build.yml
```diff
@@ -0,0 +1,254 @@
...
+                    tags: |
+                        type=pep440,pattern={{version}}
+                        type=pep440,pattern={{major}}.{{minor}}
+                        type=pep440,pattern={{major}}
+                        type=sha,prefix=
+                        type=edge
+                    labels: |
+                        org.opencontainers.image.title=resotobase
+                        org.opencontainers.image.description=Resoto base image
+                        org.opencontainers.image.vendor=Some Engineering Inc.
```
TheCatLady, line 60: Nice!  We should also add `org.opencontainers.image.created`, `org.opencontainers.image.revision`, `org.opencontainers.image.source`, `org.opencontainers.image.url`, and `org.opencontainers.image.version`.

lloesche: This is just the list of labels we override. The meta action adds the rest. We're currently setting:
```
      "org.opencontainers.image.title": "resotobase",
      "org.opencontainers.image.description": "Resoto base image",
      "org.opencontainers.image.url": "https://github.com/someengineering/resoto",
      "org.opencontainers.image.source": "https://github.com/someengineering/resoto",
      "org.opencontainers.image.version": "8d402ef",
      "org.opencontainers.image.created": "2022-01-20T18:05:52.062Z",
      "org.opencontainers.image.revision": "8d402ef44405a507eda6b5f1663d74d55d0afb08",
      "org.opencontainers.image.licenses": "Apache-2.0",
      "org.opencontainers.image.vendor": "Some Engineering Inc."
```

lloesche:
There's not much to parallelise. The base image needs to be build first for the others to be able to base `FROM` it. Building the baseimage takes 98% of the time.

.github/workflows/split-docker-build.yml
```diff
@@ -0,0 +1,254 @@
...
+                    cache-to: type=local,dest=/tmp/.buildx-cache-new
+
+            -   name: Build and push resotocore Docker image
+                if: github.event_name != 'pull_request'
+                uses: docker/build-push-action@v2
+                with:
+                    context: .
+                    file: ./Dockerfile.resotocore
+                    platforms: ${{ steps.platform.outputs.targets }}
+                    push: ${{ github.event_name != 'pull_request' }}
```
TheCatLady, line 198: This is redundant with the `if` condition specified above.  (Same with the below steps.)

.github/workflows/split-docker-build.yml
```diff
@@ -0,0 +1,254 @@
...
+                    build-args: |
+                        IMAGE_TAG=${{ steps.sha.outputs.short }}
+                    tags: ${{ steps.coremeta.outputs.tags }}
+                    labels: ${{ steps.coremeta.outputs.labels }}
+                    cache-from: type=local,src=/tmp/.buildx-cache
+                    cache-to: type=local,dest=/tmp/.buildx-cache-new
```
TheCatLady, line 204: This will overwrite the cache, and I think we only need to cache for the base image.

.github/workflows/docker-build.yml
```diff
@@ -46,6 +46,14 @@ jobs:
...
                 uses: docker/setup-buildx-action@v1
 
+            -   name: Cache Docker layers
+                uses: actions/cache@v2
+                with:
+                    path: /tmp/.buildx-cache
+                    key: ${{ runner.os }}-buildx-${{ github.sha }}
+                    restore-keys: |
+                        ${{ runner.os }}-buildx-
+
```
TheCatLady, line 56: Need to use a unique key, different than the one for the other workflow.

.github/workflows/split-docker-build.yml
lloesche: Yea I still want to get rid of that `if`. Just not sure how I'll do it exactly. It would be good to ensure these steps succeed, even on PRs. But for that we'd have to store the resotobase image, at least temporarily. I'm thinking of something like on PR storing the baseimage with the full hash as tag and then referencing that. Right now the other Dockerfiles are essentially untested until we merge the PR to main which is not ideal. It's okay for the moment as they are only 3 lines and not very error prone but that might change in the future.

TheCatLady: We would then need to push the base image to some repository anytime there are changes to a PR, though, if we want to build additional images off of them.  We could maybe create a new private package that isn't visible outside of our org?

.github/workflows/docker-build.yml
```diff
@@ -46,6 +46,14 @@ jobs:
                 id: buildx
                 uses: docker/setup-buildx-action@v1
 
+            -   name: Cache Docker layers
```
TheCatLady, line 49:
```suggestion
            -   name: Cache Docker layers
                if: github.event_name != 'pull_request'
```
Don't allow PRs to break the cache.

TheCatLady: (Additional changes below required as well.)

lloesche: But don't we want the cache to be updated by PR so that when the PR is merged to main the image layers are already in cache?

.github/workflows/split-docker-build.yml
lloesche: Why do we need to push the image? Isn't the build cache local and independent of any push to a remote repo?

.github/workflows/docker-build.yml
TheCatLady: Hm yeah, I suppose.  But then anyone could open a PR and break the cache?  Do we have actions configured to run for any user, or is approval required for first-time contributors?

lloesche: Approval is required

PR #554, opened 2022-01-19
Deploy API docs to GH Pages

TheCatLady:
Generates and deploys Swagger UI to GitHub Pages, so that we can link to it from the docs rather than instructing users to view it from their own installs.


lloesche (approved):
lgtm 🚢

PR #559, opened 2022-01-20
[resotocore] Use proper time units

aquamatthias:
So far, we used https://github.com/timwedde/durations_nlp for duration parsing, which uses single letter units as short names, which is not standard.

This PR uses durations with proper time units.
See: https://en.wikipedia.org/wiki/Unit_of_time 
Following time units are supported:

|short | long |
|------|-----|
yr | year
mo | month
d | day
h | hour
min | minute
s | second

The parts can be combined with an empty string, comma, or term.

Examples:

```
5d
5d3h12min
5 days and 12 hours
```

### DIFF
--- a/resotocore/core/durations.py
+++ b/resotocore/core/durations.py
@@ -1,0 +1,58 @@
+import operator
+from datetime import timedelta
+from functools import reduce
+import re
+from itertools import chain
+from typing import Union, List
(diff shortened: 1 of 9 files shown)


lloesche (approved):
lgtm 🚢

PR #564, opened 2022-01-21
[resotocore] Remove section aware query API

aquamatthias:
Allow the normal query API to define the section as query parameter to get the same functionality.


lloesche (approved):
lgtm 🚢

PR #568, opened 2022-01-21
[resotocore] Create user/database with default setup

aquamatthias:
Boot process:
- start to connect with username/password
- if this is not allowed
   - try to access the system database with default settings (username=root, password=None)
   - in case this is not allowed, give up
   - otherwise try to create user and database
   - in case this is not allowed, give up
- try to connect again

### DIFF
--- a/.github/workflows/resotocore_coverage.yml
+++ b/.github/workflows/resotocore_coverage.yml
@@ -24,7 +24,7 @@ jobs:
       arangodb:
         image: arangodb:3.8.3
         env:
-          ARANGO_NO_AUTH: 1
+          ARANGO_ROOT_PASSWORD: ""
         ports:
           - 8529:8529
     steps:
--- a/.github/workflows/resotocore_coverage.yml
+++ b/.github/workflows/resotocore_coverage.yml
@@ -24,7 +24,7 @@ jobs:
       arangodb:
         image: arangodb:3.8.3
         env:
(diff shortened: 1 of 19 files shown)


lloesche:
This is good and works well in my tests. Let's add two more cases:
- if you can authenticate with root and no password, after creating the database and user, change the root password to the same as `--graphdb-password`
- - add a flag that turns off this behaviour `--graphdb-bootstrap-do-not-secure`
- If the user/password or database does not exist and root/no-password does not work try root/`--graphdb-password`.

I feel like this is better default behaviour than just leaving the system open with a root user that has no password.

aquamatthias:
I like the idea, that the user can just choose a password, that is used to secure the database automatically (not only for the created database user, but also the root account).

I am not a fan of using a default password for the root account. The lazy user probably did not read why this happens and is not able to log into the database any longer.

So I would propose to implement on top:
In case there is a default root account and the configured db does not exist,  we also use the provided password to change the root account. A vanilla DB can be used and is secured after the first run with a chosen password.

aquamatthias:
Side note: while adding tests, I realized the behavior of ArangoDB is different on macOS and Linux.
On macOS: I get an authentication error, while on Linux I get a database not found error.

aquamatthias:
If we need to refine this behavior more, it can happen in a separate PR.

PR #572, opened 2022-01-24
[resotolib/resotoworker] Allow graph to have multiple directed edges of different types

lloesche:


resotolib/resotolib/graph/__init__.py
```diff
@@ -166,16 +199,81 @@ def add_edge(self, src, dst, **attr):
...
+            edge_type = EdgeType.default
+        edges = []
+        for edge in self.edges(keys=True):
+            if len(edge) == 3:
+                key: EdgeKey = edge[2]
+                if key.edge_type == edge_type:
+                    edges.append(edge)
+        return self.edge_subgraph(edges)
+
+    def is_dag(self) -> bool:
```
aquamatthias, line 261: Only naming: readers might think that if this returns true the whole multigraph is a dag, which is not true.

aquamatthias (commented):
One little nit - otherwise lgtm. thanks.

aquamatthias (approved):

resotolib/resotolib/graph/__init__.py
lloesche: good point, renamed

PR #575, opened 2022-01-24
[github] Add PR template & update issue templates

TheCatLady:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+<!-- Please describe the changes included in this PR here. -->
+
+# To-Dos
+
+<!-- Before submitting this PR, please lint and build your changes locally. -->
+<!-- Add an 'x' between the brackets to mark each checkbox as checked. -->
+
+- [ ] Lint and test with `tox`
+
```
aquamatthias, line 11:
```suggestion
- [ ]  PR includes a test that will fail if the code is reverted, but the test is not.

```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+# Issues Fixed
+
+<!-- If this PR will fix/resolve an open issue on the repository, please reference it below. -->
+<!-- Otherwise, please delete this section. -->
+
+- Fixes #XXXX
```
aquamatthias, line 17: Please move the example into the html comment.
So nothing is rendered if lazy people forget to delete this part.

aquamatthias (commented):
2 small additions - otherwise looks fine to me.

.github/PULL_REQUEST_TEMPLATE.md
TheCatLady: The heading would need to be moved as well, in that case. I don't think there's a good way to handle this until GitHub supports pull request forms.

.github/PULL_REQUEST_TEMPLATE.md
TheCatLady: How about:

```
- [ ] Test coverage for new code
```

Your suggestion is a little difficult to understand.

.github/PULL_REQUEST_TEMPLATE.md
TheCatLady: There is also no harm in the user not deleting this section. But there is a lot of benefit to making it easy for the user to link related issues. :)

.github/PULL_REQUEST_TEMPLATE.md
aquamatthias: It is probably related but different.
If it is too complex to understand - let's not add it.

.github/PULL_REQUEST_TEMPLATE.md
aquamatthias: I prefer to hint at what a good PR description could look like - I assume #XXXX will get fixed often.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
+# Description
+
+<!-- Please describe the changes included in this PR here. -->
+
+# To-Dos
+
+<!-- Before submitting this PR, please lint and build your changes locally. -->
+<!-- Add an 'x' between the brackets to mark each checkbox as checked. -->
+
+- [ ] Test coverage for new code
```
aquamatthias, line 10: Somewhat unclear to me. Maybe something like:

- [ ] Added or changed functionality is covered by tests

wdyt?

aquamatthias (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
TheCatLady: How about:

```
- [ ] Test coverage added for new or updated functionality
```

aquamatthias: Like it

aquamatthias (approved):

PR #599, opened 2022-01-30
[docker] Initial docker-compose.yml

lloesche:
# Description

Adds a compose file using the new split up Docker images

# Code of Conduct


docker-compose.yaml
```diff
@@ -0,0 +1,60 @@
...
+    image: ghcr.io/someengineering/resotocore
+    depends_on:
+      - graphdb
+    ports:
+      - "8900:8900/tcp"
+    environment:
+      - PSK
+      - RESOTOCORE_GRAPHDB_SERVER=http://graphdb:8529
+      - RESOTOCORE_GRAPHDB_PASSWORD=changeme
+    restart: always
```
TheCatLady, line 23:
```suggestion
    restart: on-failure[:max-retries]
```
It may be better to set a number of retries to prevent a restart loop.

docker-compose.yaml
```diff
@@ -0,0 +1,60 @@
...
+    stop_grace_period: 2m
+  resotoworker: 
+    image: ghcr.io/someengineering/resotoworker
+    depends_on:
+      - resotocore
+    ports:
+      - "9956:9956/tcp"
+    environment:
+      - PSK changeme
```
TheCatLady, line 32: Copy/paste error?

TheCatLady (commented):
Can use extension fields so you there is not so much repetition/duplication of env vars, etc. in the different services: https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

PR #613, opened 2022-02-02
[resoto][feat] Introduce poetry

meln1k:
# Description

Replace the setuptools/tox with poetry/nox.

Motivation: setuptools does not capture the exact version of transitive dependencies, which can lead to unstable builds. Poetry can do it, and nox can integrate with poetry to run deterministic tests.

Steps needed for completing the PR:
- [x] Move resotolib to poetry
- [x] Move resotocore to poetry
- [x] Move resotoshell to poetry
- [x] Move resotometrics to poetry
- [x] move resotoworker to poetry
- [x] Move plugins to poetry
- [x] Adjust the docker images to work with poetry
- [x] Adjust the setup_venv.sh script
- [x] Adjust the bump version script
- [x] Adjust the docs
- [x] Adjust the github template

# To-Dos

# Code of Conduct


TheCatLady:
Nice!  I know this is still a WIP, but don't forget to also update the [`someengineering/resoto` PR template](https://github.com/someengineering/resoto/blob/main/.github/PULL_REQUEST_TEMPLATE.md) as well as the [Resoto component contribution guide](https://resoto.com/docs/contributing/components/) (will need to submit a PR to `someengineering/resoto.com` to update docs).

Dockerfile.resotobase
```diff
@@ -80,50 +80,51 @@ RUN mkdir -p /build-python
...
 # Build resoto plugins
 COPY plugins /usr/src/plugins
 WORKDIR /usr/src
 RUN if [ "X${TESTS:-false}" = Xtrue ]; then . /usr/local/resoto-venv-python3/bin/activate && find plugins/ -name tox.ini | while read toxini; do cd $(dirname "$toxini") && tox && cd - || exit 1; done; fi
-RUN . /usr/local/resoto-venv-python3/bin/activate && find plugins/ -maxdepth 1 -mindepth 1 -type d -print0 | xargs -0 python -m pip wheel -w /build-python -f /build-python
-
-# Install all wheels
-RUN . /usr/local/resoto-venv-python3/bin/activate && python -m pip install -f /build-python /build-python/*.whl
-RUN . /usr/local/resoto-venv-pypy3/bin/activate && pypy3 -m pip install -f /build-pypy /build-pypy/*.whl
+RUN for plugin_dir in $(find plugins/ -maxdepth 1 -mindepth 1 -type d); do \
```
lloesche, line 123: A for loop won't work here because some of the plugins depend on another. E.g. the cleanup_aws_* depend on aws. If it works it's only through luck, because it got to the dependent plugin first.

meln1k: good point. The logic here is the same as it was before when the output of the `find` command was used. I will create a ticket to rewrite this part.

.github/workflows/check_pr_plugin_aws.yml
```diff
@@ -30,19 +30,20 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/.cache/pip
```
aquamatthias, line 32: According to https://python-poetry.org/docs/configuration/#cache-dir poetry uses this path: `~/.cache/pypoetry`.
Maybe we should define `~/.cache` here.
Same is true for all other python GitHub actions.

.github/workflows/check_pr_plugin_cleanup_aws_loadbalancers.yml
```diff
@@ -30,19 +30,20 @@ jobs:
...
-          pip install --upgrade --editable resotolib/
-          pip install tox wheel flake8
+          pip install poetry nox
 
       - name: Run tests
         working-directory: ./plugins/cleanup_aws_loadbalancers
-        run: tox
+        run: |
+          nox
+          nox -rs coverage_ci
```
aquamatthias, line 46: Instead of 2 separate runs - can we define the list of targets that should be executed?
Same is true for all other GitHub actions.

aquamatthias (dismissed):
Please find my last 2 comments. Otherwise LGTM. Thanks.

aquamatthias (approved):
Thanks!

PR #637, opened 2022-02-08
[resotocore][fix] Improve error message in case of timeout.

aquamatthias:
# Description

Instead of throwing `queue.Empty`, raise a more meaningful exception.


MrMarvin (approved):
👍

lloesche (approved):
lgtm 🚢

PR #640, opened 2022-02-08
[resotocore][feat] successors,predecessors,ancestors, descendants useedge_type from env

aquamatthias:
# Description

The edge type can be either defined as an argument or as env var.
The argument takes always precedence over the env var.

# To-Dos


lloesche (approved):
lgtm 🚢

PR #643, opened 2022-02-09
[resotocore][fix] Fix aggregation functions using the same variable

aquamatthias:
# Description

Queries that use the same variable in different aggregation functions did not work properly.
```
> query is(instance) | aggregate sum(1) as count,  sum(instance_cores) as cores, sum(instance_memory) as memory, max(instance_memory) as max_mem
```

# To-Dos


lloesche (approved):
lgtm 🚢

PR #644, opened 2022-02-09
[plugin gcp][feat] Add GKE clusters resource

MrMarvin:
# Description

Adding a GKE cluster resource with collector to the GCP plugin. This allows to collect `gke_clusters`. 
Please note that we are not collecting GKE nodepools yet and therefore do not link any instances.

```
> query is(gcp_gke_cluster) | head 3
kind=gcp_gke_cluster, id=b1256ccc4e044930b78633db406ccbec65792471847f418e979667415d30aa23, name=us-east1-fast-demo-9ccc7f42-gke, age=8mo23d, cloud=gcp, account=acme-labs
kind=gcp_gke_cluster, id=1ccc7052286042049a861bdd7afbe73cd75e0649f6e8477d9c30ec3744051e42, name=test-cluster-one, age=4mo25d, cloud=gcp, account=acme-labs
kind=gcp_gke_cluster, id=f0a9caae11dd416d961814188af5e8198c4b5714746e424ca6abff3204706042, name=marv-private-terraform-test-2, age=6d12h, cloud=gcp, account=acme-labs
```

# To-Dos
- [X] Lint and test with `tox`

The GCP plugin ships with practically no tests. Manual tests collections confirmed it is working as expected, as well as not breaking other resource type collections.

Additionally I'm unsure how to proceed with the documentation for the new resource. Is that something you folks can help me with?

# Code of Conduct

### DIFF
--- a/plugins/gcp/resoto_plugin_gcp/collector.py
+++ b/plugins/gcp/resoto_plugin_gcp/collector.py
@@ -553,7 +553,7 @@ class GCPProjectCollector:
         successors: List = None,
         predecessors: List = None,
         client_kwargs: Dict = None,
-        client_nested_callables: Type[str] = [],
+        client_nested_callables: List[str] = None,
         resource_kwargs: Dict = None,
         paginate_subitems_name: str = None,
         post_process: Callable = None,
@@ -594,6 +594,8 @@ class GCPProjectCollector:
             paginate_subitems_name = client_method_name
         if client_kwargs is None:
             client_kwargs = {}
+        if client_nested_callables is None:
+            client_nested_callables = []
         if resource_kwargs is None:
             resource_kwargs = {}
         if successors is None:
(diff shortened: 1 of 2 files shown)


plugins/gcp/resoto_plugin_gcp/collector.py
```diff
@@ -547,6 +553,7 @@ def collect_something(
         successors: List = None,
         predecessors: List = None,
         client_kwargs: Dict = None,
+        client_nested_callables: Type[str] = [],
```
lloesche, line 556: Should this be `List[str]`?

If `client_nested_callables` should default to an empty list the better pattern to use is
```
client_nested_callables: List[str] = None,
if client_nested_callables is None:
    client_nested_callables = []
```

Otherwise every instance of this class will share the same reference to that empty list and if one manipulates it it changes in all the other object instances as well.

lloesche:
@MrMarvin this is great, thank you!
re: docs, the plugins are currently not documented at all. @TheCatLady is doing a large documentation revamp which will also include cloud specific instructions. So for this PR we're good on that front.

plugins/gcp/resoto_plugin_gcp/collector.py
MrMarvin: Makes sense, thanks!

lloesche (approved):
lgtm 🚢

PR #666, opened 2022-02-16
[resotoshell][fix] Use lru_cache instead of cache

aquamatthias:
# Description

The cache directive is only available in python 3.9 or later.

### DIFF
--- a/resotoshell/resotoshell/__main__.py
+++ b/resotoshell/resotoshell/__main__.py
@@ -3,7 +3,7 @@ import pathlib
 import re
 import shutil
 import sys
-from functools import cache
+from functools import lru_cache
 from threading import Event
 from typing import Dict, Union, Any, Optional
 from urllib.parse import urlencode, urlsplit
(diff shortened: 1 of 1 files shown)


resotoshell/resotoshell/__main__.py
```diff
@@ -194,7 +194,7 @@ def decode(value: Union[str, bytes]) -> str:
             handle_result(part, num == 0)
 
 
-@cache
+@lru_cache()
```
meln1k, line 197: This is not 1:1 replacement. To follow the `@cache` behaviour it needs to be `lru_cache(maxsize=None)`

meln1k, line 197:
```suggestion
@lru_cache(maxsize=None)
```

aquamatthias: Since we only cache exactly one element - it should not matter?

meln1k: not really, except it is faster due to not having the cache size checks

lloesche (approved):
lgtm 🚢

PR #672, opened 2022-02-16
[resotocore][feat] Add a csv output option to the list command

### LINKED ISSUE #653: [resotocore] add CSV and markdown table output format
### Motivation

From Discord discussion:
```
@isme: CSV output would certainly be useful because it's a format that pretty much every analysis tool supports.

But sometimes, as in the above example, I just use CSV as a "staging" format so that I can render a table using csvlook. A table is often the most readable way to present results.

So if resh also had a table format it would save a lot of manual labor. You'd get bonus points if the table output was in Github syntax because then you could immediately paste it into a Markdown document or Github issue.
```

### Desired Behavior

Have a CSV format output, including header, as well as markdown pre-formated table output.

### Additional Context

_No response_

meln1k:
# Description

Added a new --csv flag for the list command. Once specified, query will result in CSV formatted output.

# To-Dos

# Issues Fixed

- Fixes #653 

# Code of Conduct


resotocore/resotocore/cli/command.py
```diff
@@ -1983,10 +1989,32 @@ def fmt_json(elem: Json) -> JsonElement:
             else:
                 return elem
 
-        def fmt(elem: JsonElement) -> JsonElement:
-            return fmt_json(elem) if isinstance(elem, dict) else str(elem)
+        async def fmt_csv(elem: Json, idx: int) -> AsyncIterator[JsonElement]:
+            def to_csv_string(lst: List[Any]) -> str:
```
aquamatthias, line 1993: Mmh idx is actually not required - you could do something like this:
```python
async def csv_stream(in_stream: Stream) -> AsyncIterator[JsonElement]:
   yield "header"
   async for elem in in_stream:
       yield to_csv(elem)
```
Wouldn't this be simpler?

resotocore/resotocore/cli/command.py
```diff
@@ -1983,10 +1989,32 @@ def fmt_json(elem: Json) -> JsonElement:
...
+                output = io.StringIO()
+                dialect = csv.unix_dialect()
+                writer = csv.writer(output, dialect=dialect, quoting=csv.QUOTE_MINIMAL)
+                writer.writerow(lst)
+                return output.getvalue().rstrip()
```
aquamatthias, line 1998: This will create a new string/writer etc for every element...
You can create this once for the whole stream and then call:
```python
output.truncate()
writer.writerrow(let)
output.getvalue()
```
for every element, reusing the buffer.

resotocore/resotocore/cli/command.py
```diff
@@ -1863,6 +1864,13 @@ class ListCommand(CLICommand, OutputTransformer):
 
       *Example*: `path.to.property as prop1`.
 
+
+    ## Options
+
+    - --csv [optional]: if set, the output will be formatted as CSV. Can't be used together with --markdown.
+
+    - --markdown [optional]: if set, the output will be formatted as Markdown. Can't be used together with --csv.
```
aquamatthias, line 1872:
```suggestion
    - `--csv` [optional]: format the output as CSV. Can't be used together with `--markdown`.

    - `--markdown` [optional]: format the output as Markdown table. Can't be used together with `--csv`.
```

resotocore/resotocore/cli/command.py
```diff
@@ -1889,6 +1897,20 @@ class ListCommand(CLICommand, OutputTransformer):
...
+
+    # Properties that do not exist will be printed as empty values when using csv or markdown output.
+    > query is(aws_ec2_instance) limit 3 | list --csv kind as a, name as b, does_not_exist
+    a,b,does_not_exist
+    aws_ec2_instance,sun,
+    aws_ec2_instance,moon,
+    aws_ec2_instance,star,
+
+    > query is(aws_ec2_instance) limit 3 | list --markdown kind as a, name as b, does_not_exist
+    |a               |b   |does_not_exist|
```
aquamatthias, line 1909: Consider to us a different example (is(aws_ec2_instance) will always yield the same kind)
```
> search is(instance) limit 3 | list instance_cores as cores, name, does_not_exist
```

resotocore/resotocore/cli/command.py
```diff
@@ -1889,6 +1897,20 @@ class ListCommand(CLICommand, OutputTransformer):
...
+    |----------------|----|--------------|
+    |aws_ec2_instance|sun |None          |
+    |aws_ec2_instance|moon|None          |
+    |aws_ec2_instance|star|None          |
```
aquamatthias, line 1913: `None` is actually wrong. A non existent value should be printed as `null`.

aquamatthias (commented):
LGTM - Thanks. Added some small refinements.

aquamatthias (approved):
LGTM- Thanks!

PR #680, opened 2022-02-21
[resotocore][feat] Show ck mascot on help

aquamatthias:
# Description

image

Do not show the logo if color is not supported.


meln1k (dismissed):
Awesome. Thanks!

meln1k:
What do you think about having the resoto cli version to the right from the image? so that the empty space is filled with something

lloesche (approved):
lgtm 🚢

PR #682, opened 2022-02-22
[plugins/aws][feat] Add support for Cloudformation Stack Sets

### LINKED ISSUE #651: [plugins/aws] Add CloudFormation stack set resource
### Motivation

I use CloudFormation stack sets a lot in one project.

AWS Config doesn't have AWS::CloudFormation::StackSet in its [resource schema](https://github.com/awslabs/aws-config-resource-schema) yet, so I have to use [custom code](https://github.com/iainelder/aws-stack-set-snapshot) to get a snapshot of them.

Each stack set needs multiple API calls to fully describe it.

* DescribeStackSet
* ListStackInstances
* ListStackSetOperations

There are perhaps even more API calls to be made, but I haven't needed more detail than that so far.

### Desired Behavior

That when I have 5 stack sets in the account and I run a query like this:

```
query is(aws_resource) | count kind
```

Then I get a result like this in the output:

```
aws_cloudformation_stack_set 5
```

I'm not sure whether it should be spelled "stack_set" or "stackset". I'll leave that to your discretion.

I'm not sure what the data model or queries for the stack instances and operations would look like. I'm still new to Resoto.

### Additional Context

_No response_

lloesche:
# Description

Add support for Cloudformation Stack Sets. This is part 1 of a 2 part PR. Part 2 is for the commented code that requires support in resotocore for creating edges between accounts, after all accounts have been collected. We could already implement this on the collector level in the form of a deferred_connection but that would only work for `--graph-merge-kind cloud` and break if a user collects two different accounts on two different workers.

# Issues Fixed

- Fixes #651

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/accountcollector.py
+++ b/plugins/aws/resoto_plugin_aws/accountcollector.py
@@ -2203,10 +2203,12 @@ class AWSAccountCollector:
         session = aws_session(self.account.id, self.account.role)
         client = session.client("cloudformation", region_name=region.id)
 
-        response = client.list_stack_sets()
+        response = client.list_stack_sets(Status="ACTIVE")
         stack_sets = response.get("Summaries", [])
         while response.get("NextToken") is not None:
-            response = client.list_stack_sets(NextToken=response["NextToken"])
+            response = client.list_stack_sets(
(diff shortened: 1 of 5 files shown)


aquamatthias (dismissed):
:shipit:

PR #689, opened 2022-02-24
[resotocore][fix] Send ModelInfo event after one hour uptime

aquamatthias:
# Description

Only sent this event after one hour. Installations that will start and stop under one hour will not send this event.


lloesche (approved):
lgtm 🚢

PR #691, opened 2022-02-24
[resotocore][fix] backup/restore did not recreate the graph

aquamatthias:
# Description

I realized that a backup -> restore cycle did restore all data but not the graph description itself.
This information is considered internal and needs to be included explicitly.


lloesche (approved):
lgtm 🚢

PR #692, opened 2022-02-24
[resotocore][fix] ancestors should not hold a reference to the current node

aquamatthias:
# Description

Currently, the account node lists the current node under `/ancestors` as the account node.


lloesche (approved):
lgtm 🚢

PR #701, opened 2022-03-08
[resotocore][fix] Only render edges where the nodes have been visited

aquamatthias:
# Description

Edges are deferred when the nodes are not visited yet.
In the end, a final check is required to ensure both nodes have been visited.


lloesche (approved):
lgtm 🚢

PR #707, opened 2022-03-14
[resotocore][feat] Allow singular and plural form for cli commands

aquamatthias:
# Description

Allow `workflow` and `workflows`, `tag` and `tags` etc.


lloesche (approved):
yeay 🚢

PR #710, opened 2022-03-16
[resotocore][feat] Only update config on change + revision header

aquamatthias:
# Description

Config is only ipdayted if changed - also the event is also only emitted in this case.
The revision number comes as part of GET/PUT/PATCH requests.


lloesche (approved):
yeay 🚢

PR #711, opened 2022-03-16
[resotolib/resotoworker/resotometrics] Dynamic config loading and updates

lloesche:
# Description

Add support for runtime config loading and updates using the core's `/config` API.

# Code of Conduct


plugins/example_collector/test/test_args.py
```diff
@@ -6,4 +6,6 @@ def test_args():
     arg_parser = get_arg_parser()
     ExampleCollectorPlugin.add_args(arg_parser)
     arg_parser.parse_args()
-    assert ArgumentParser.args.example_region is None
+
+
+#    assert ArgumentParser.args.example_arg is None
```
aquamatthias, line 13: left over?

plugins/example_collector/test/test_config.py
```diff
@@ -0,0 +1,11 @@
...
+from resoto_plugin_example_collector import ExampleCollectorPlugin
+
+
+def test_config():
+    config = Config("dummy", "dummy")
+    ExampleCollectorPlugin.add_config(config)
+    Config.init_default_config()
+
+
+#    assert Config.example.region is None
```
aquamatthias, line 11: left over?

plugins/tagvalidator/resoto_plugin_tagvalidator/config.py
```diff
@@ -0,0 +1,85 @@
+from dataclasses import dataclass, field
+from typing import ClassVar, Dict, Union, List
+from resotolib.utils import parse_delta
+
+
+default_config = {
```
aquamatthias, line 6: Looks like this could be a typed config?

resotometrics/resotometrics/config.py
```diff
@@ -0,0 +1,66 @@
...
+    kind: ClassVar[str] = "metric"
+    help: str = field(metadata={"description": "Metric help text"})
+    search: str = field(metadata={"description": "Aggregation search to run"})
+    type: str = field(metadata={"description": "Type of metric (gauge or counter)"})
+
+
+def _load_default_metrics() -> Dict[str, Metric]:
+    default_metrics = {}
+    local_path = os.path.abspath(os.path.dirname(__file__))
+    default_metrics_file = f"{local_path}/default_metrics.yaml"
```
aquamatthias, line 23: Can we add a test that ensures, that this file can be read?

aquamatthias (commented):

plugins/example_collector/test/test_config.py
lloesche: No, just an example of how to access the Config without actually adding any examples to the user's config.

plugins/example_collector/test/test_args.py
lloesche: Nope, same as with the Config, this is an example of how you would access args had they been added to the Argparser. But I commented the code that adds them so that normal users don't have `--example-arg` listed as one of the worker args.

resotometrics/resotometrics/config.py
lloesche: I mean, if the file can't be read it'll throw an appropriate exception. If the contents of the file are wrong/missformated, the core will fail the validation on "config save" since all of the file's content ends up in fields that are strictly typed.

plugins/tagvalidator/resoto_plugin_tagvalidator/config.py
lloesche: Yes, it could, it should.

aquamatthias (approved):
looks awesome to me - thanks!

PR #758, opened 2022-04-05
[resotocore][fix] Define default host in docker env

aquamatthias:
# Description

- read the git commit hash created by docker build
- detect if running inside docker
- adjust the default hosts value according to docker detection


meln1k (approved):
LGTM. thanks!

PR #759, opened 2022-04-05
[resoto][feat] Dot Rendering script

meln1k:
# Description

Added a script that can render the graph in a dot format.

# To-Dos

# Code of Conduct

### DIFF
--- a/resotocore/tools/render_dot.py
+++ b/resotocore/tools/render_dot.py
@@ -11,9 +11,11 @@ from resotocore.util import (
     count_iterator,
 )
 from resotocore.model.resolve_in_graph import NodePath
+from resotocore.util import uuid_str, utc
 from collections import defaultdict
 import re
 from dataclasses import dataclass
+from posthog import Client
 
 JsonElement = Union[str, int, float, bool, None, Mapping[str, Any], Sequence[Any]]
 
@@ -61,7 +63,7 @@ def generate_icon_map():
     security = "Arch_Security-Identity-Compliance"
     containers = "Arch_Containers"
     networking = "Arch_Networking-Content-Delivery"
-    size = "16"
+    size = "32"
     prefix_amazon = "Arch_Amazon"
     prefix_aws = "Arch_AWS"
     prefix = "Arch"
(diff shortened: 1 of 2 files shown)


resotocore/tools/render_dot.py
```diff
@@ -0,0 +1,212 @@
...
+            distinct_id=system_id,
+            event="dot-rendering-script-run",
+            properties={"run_id": run_id},  # type: ignore
+            timestamp=now,
+        )    
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("json_dump", help="Resoto json dump")
```
aquamatthias, line 200: Just thinking if it would be much easier, if the user could call this script with a query directly?
This script could call the search/graph endpoint and get the resulting json.
wdyt?

resotocore/tools/render_dot.py
```diff
@@ -0,0 +1,212 @@
...
+    "volume": ResourceKind.VOLUME,
+    "image": ResourceKind.IMAGE,
+    "firewall": ResourceKind.FIREWALL,
+    "kubernetes_cluster": ResourceKind.K8S_CLUSER,
+    "network": ResourceKind.NETWORK,
+    "load_balancer": ResourceKind.LOAD_BALANCER,
+}
+
+
+def parse_kind(kind: str) -> Optional[ResourceKind]:
```
aquamatthias, line 63: It would be straightforward to fetch the model from the core.
This way you could use the base kinds like instance, volume etc. as fallback and specific icons if available (e.g. aws).

aquamatthias (commented):
Thanks - I added 2 proposals to make life easier. Let's discuss.

resotocore/tools/render_dot.py
meln1k: We decided to postpone this for now.

resotocore/tools/render_dot.py
```diff
@@ -0,0 +1,334 @@
...
+        yield f' subgraph "{account}" {{'
+        for uid in uids:
+            yield f'    "{uid}"'
+        yield " }"
+
+    yield "}"
+
+
+def ensure_assets():
+    if not os.path.exists("Assets"):
```
lloesche, line 196: how about `~/.resoto/cache/assets/`? That way it doesn't pollute the user's working directory and is always in the same directory, not relative to where the user currently is.

lloesche: Even if the directory exists we should check that it contains what we expect it to contain.

lloesche (dismissed):

aquamatthias (approved):

aquamatthias (approved):

resotocore/tools/render_dot.py
meln1k: hmm, that's a bit more complicated. I will create a directory with a more unique name for now.

PR #769, opened 2022-04-06
[resotolib][fix] Fix support for types in argparse env handling

aquamatthias:
# Description

See title.

### DIFF
--- a/resotolib/resotolib/args.py
+++ b/resotolib/resotolib/args.py
@@ -135,20 +135,14 @@ def convert(value: Any, type_goal: Union[type, Callable]) -> Any:
         return value
     elif isinstance(type_goal, type):
         try:
-            if type_goal is str:
-                return str(value)
+            if type_goal in (str, int, float, complex):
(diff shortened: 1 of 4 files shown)


resotolib/resotolib/args.py
```diff
@@ -126,16 +127,29 @@ def get_arg_parser(
...
-        pass
-    else:
-        value = converted_value
-    return value
+def convert(value: Any, type_goal: Union[type, Callable]) -> Any:
+    if type_goal is NoneType:
+        return value
+    elif isinstance(type_goal, type):
+        try:
+            if type_goal is str:
```
lloesche, line 135: Could this be turned into
```
            if type_goal in (str, int, float, complex):
                return type_goal(value)
            elif type_goal is bool:
                return value.lower() in ("true", "1", "yes")
```

aquamatthias: 👍

lloesche (approved):
lgtm 🚢

PR #770, opened 2022-04-06
[resotocore][fix] Allow simple and list values as override parameters

aquamatthias:
# Description

Allows defining simple and list override parameters.
Example:
```
resotocore --override resotocore.api.web_hosts=localhost,case resotocore.api.web_port=12345
```
Would bind to localhost and case with port 12345


lloesche (approved):
lgtm 🚢

PR #794, opened 2022-04-19
[plugins][fix] Rename protect_snowflakes -> protector and make plugins restart on enable

lloesche:
# Description

This PR:
- renames plugin `protect_snowflakes` to `protector`
- sets the restart flag for all plugins enable option
- reverts the change that made ActionPlugins run as Process instead of Thread

# Code of Conduct


resotolib/resotolib/baseplugin.py
```diff
@@ -133,10 +134,10 @@ def action_processor(self, message: Dict) -> None:
...
     def run(self) -> None:
         try:
-            ArgumentParser.args = self._args
-            resotolib.config._config = self._config
-            setup_logger("resotoworker")
-            resotolib.signal.initializer()
+            # ArgumentParser.args = self._args
+            # resotolib.config._config = self._config
+            # setup_logger("resotoworker")
+            # resotolib.signal.initializer()
```
aquamatthias, line 140: can be removed?

aquamatthias (approved):

resotolib/resotolib/baseplugin.py
lloesche: I want to eventually go back to using Process so I'm keeping it here for reference of what's needed then.

PR #831, opened 2022-05-03
[resotoshell][feat] select mode if stdin is not a tty

aquamatthias:
# Description

Detect if stdin is a tty and read from stdin if not.
This allows:
```sh
echo "search is(instance) limit 3" | resh
```
without providing extra options.


lloesche (approved):
Nice! 🚢

PR #845, opened 2022-05-09
[docker][fix] Increase file limits for ArangoDB

aquamatthias:
# Description

- increase default file descriptor limit - much too low for arangodb
- define minimal resources for graphdb, core and worker
- define root password (empty value is swallowed in ECS)


docker-compose.yaml
```diff
@@ -27,12 +36,17 @@ services:
...
     ports:
-      - 8900:8900
+      - "8900:8900"
     environment:
       - PSK
-      - RESOTOCORE_GRAPHDB_SERVER=http://graphdb:8529
-      - RESOTOCORE_GRAPHDB_PASSWORD=changeme
     command:
+      - --graphdb-server
+      - http://graphdb:8529
```
lloesche, line 44: Should we switch this to args?
a) the password will be in plain text in the process list of the host
b) the password should come from something like a k8s secret or vault and will then be handed in via the env.
By switching from env to cli flags we're teaching a bad behaviour.

docker-compose.yaml
```diff
@@ -5,15 +5,24 @@ services:
     image: arangodb:3.8.6
     platform: linux/amd64
     environment:
-      - ARANGO_ROOT_PASSWORD=
+      - ARANGO_ROOT_PASSWORD=changeme
```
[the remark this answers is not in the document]
aquamatthias: docker swallows empty environment variables in some contexts.
Result: the arango container does not start and complains about a root password setting.
I experienced it while using the AWS ECS context and would opt to work around it.

docker-compose.yaml
aquamatthias: You are right - changed back to env var.

lloesche (approved):

PR #868, opened 2022-05-16
[resotocore][fix] Export model in correct format

aquamatthias:
# Description

API spec defined an array, but it returned a json object.

# To-Dos


lloesche (approved):
thanks 🚢

PR #891, opened 2022-05-25
[resotocore][fix] search navigation after tail

aquamatthias:
# Description

```bash
search all | tail -1 | successors
```

Did not execute and threw an errror. Now this search is possible.


meln1k (approved):

lloesche (approved):
yeay 🎉

PR #894, opened 2022-05-30
[resoto][feat] Initial DNS support

lloesche:
# Description

Initial DNS support.
This will be a multi-part PR.

The first one adds Route53 support for DNS Zones and resource records.
The second PR adds edge support between RRs for CNAMEs and in the Route53 case for ALIAS RRs.
The third PR adds delete support.
The fourth PR adds external edge support so that an A or AAAA record can point to e.g. a compute instance or load balancer.

# Code of Conduct


resotolib/resotolib/utils.py
```diff
@@ -846,3 +846,72 @@ def safe_members_in_tarfile(tarfile: TarFile) -> List:
         else:
             result.append(tar_info)
     return result
+
+
+def rrdata_as_dict(record_type: str, record_data: str) -> Dict:
```
aquamatthias, line 851: consider to move to a `dns.py` file.

aquamatthias (approved):
thanks

PR #939, opened 2022-06-20
[k8s][chore] Update Readme.

aquamatthias:
# Description

Point to main documentation.


TheCatLady:
Do we want to link directly to the Kubernetes section once it's reviewed and merged?

TheCatLady:
> If we change the structure, it is easy to break links. That is why I wanted to use a very "stable" link (could be just resoto.com, if you think this is better).

I have actually been adding server-side redirects anytime paths change. I think there is benefit to linking directly to the relevant content rather than expecting the user to search for it, but ideally we would maintain and update any links.

Whatever you decide to do here is fine, it doesn't seem like too big of a deal :)

PR #976, opened 2022-07-06
[resotocore][fix] Allow for empty Discord alert message & move to embed description

TheCatLady:
# Description

As I was editing the Discord how-to guide, I noticed that the current Discord alerts look a little unpolished:
image

We should allow for users to omit the alert message entirely if they only want to specify the title. Also, the message text should be contained within the embed object for a cleaner look (and also so that the message doesn't look like a title).

# To-Dos

# Code of Conduct

### DIFF
--- a/resotocore/resotocore/core_config.py
+++ b/resotocore/resotocore/core_config.py
@@ -181,23 +181,21 @@ def alias_templates() -> List[AliasTemplateConfig]:
     return [
         AliasTemplateConfig(
             "discord",
-            "Send result of a search to discord",
+            "Send the result of a search to Discord",
             # defines the fields to show in the message
             "jq {name:{{key}}, value:{{value}}} | "
             # discord limit: https://discord.com/developers/docs/resources/channel#embed-object-embed-limits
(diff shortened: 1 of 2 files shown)


aquamatthias:
The current `discord` command allows to define the message to show.
Current Version:
image

Your version: 
image

Having the message underneath the title makes sense.
You also removed the default message - not sure if on purpose - can you add it again?

Otherwise LGTM.

TheCatLady:
> The current `discord` command allows to define the message to show. Current Version: image
> 
> Your version: image Having the message underneath the title makes sense. You also removed the default message - not sure if on purpose - can you add it again?
> 
> Otherwise LGTM.

As mentioned in the PR description, I think we should allow for omitting the message. What are your thoughts on this?  I don't think the current default message looks/reads very well either, but that's besides the point -- we shouldn't force users to include a message?  Similar to how the title is optional.

aquamatthias:
image

`message` is now optional - the user only needs to define `title` and `webhook` only.

aquamatthias (approved):

PR #982, opened 2022-07-06
[resotocore][fix] Migrate to zoneinfo

lloesche:
# Description

Migrate to zoneinfo

Fixes
```
/Users/lukas/repo/resoto/venv-pypy/lib/pypy3.9/site-packages/apscheduler/util.py:436: PytzUsageWarning: The localize method is no longer necessary, as this time zone supports the fold attribute (PEP 495). For more details on migrating to a PEP 495-compliant implementation, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  return tzinfo.localize(dt)
```

Also adds a dependency to tzdata because of https://docs.python.org/3/library/zoneinfo.html#data-sources
Normal Linux distributions should come with a timezone db, but this prepares it to work on systems that don't come with one, and more importantly also makes it work in minimal container environments that don't come with a system timezone db.

# Code of Conduct

### DIFF
--- a/resotocore/resotocore/task/scheduler.py
+++ b/resotocore/resotocore/task/scheduler.py
@@ -1,5 +1,4 @@
 from datetime import datetime
-from zoneinfo import ZoneInfo
 from typing import Callable, Any, List
 
 from apscheduler.executors.asyncio import AsyncIOExecutor
@@ -7,6 +6,8 @@ from apscheduler.job import Job
 from apscheduler.jobstores.memory import MemoryJobStore
 from apscheduler.schedulers.asyncio import AsyncIOScheduler
 from apscheduler.triggers.cron import CronTrigger
+
+from resotolib.utils import get_local_tzinfo
 
 
 class Scheduler:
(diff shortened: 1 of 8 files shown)


resotocore/resotocore/task/scheduler.py
```diff
@@ -18,7 +18,7 @@ def __init__(self) -> None:
             # max_instances: allowed parallel instances for one job
             # misfire_grace_time: seconds after the designated runtime that the job is still allowed to be run
             job_defaults={"coalesce": True, "max_instances": 32, "misfire_grace_time": 3600},
-            timezone=utc,
+            timezone=ZoneInfo("Etc/UTC"),
```
aquamatthias, line 21: Do we want to use the local time zone here and fallback to UTC?

aquamatthias (dismissed):

resotocore/resotocore/task/scheduler.py
lloesche: yes!

meln1k (approved):

PR #992, opened 2022-07-10
[plugin/aws][feat] Add support for Lambda Functions

fernandocarletti:
# Description
Add support for Lambda Functions

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/accountcollector.py
+++ b/plugins/aws/resoto_plugin_aws/accountcollector.py
@@ -112,6 +112,10 @@ metrics_collect_instances = Summary(
     "resoto_plugin_aws_collect_instances_seconds",
     "Time it took the collect_instances() method",
 )
+metrics_collect_lambda_functions = Summary(
+    "resoto_plugin_aws_collect_lambda_functions_seconds",
+    "Time it took the collect_lambda_functions() method",
+)
 metrics_collect_keypairs = Summary(
     "resoto_plugin_aws_collect_keypairs_seconds",
     "Time it took the collect_keypairs() method",
@@ -287,6 +291,7 @@ class AWSAccountCollector:
             "internet_gateways": self.collect_internet_gateways,
             "keypairs": self.collect_keypairs,
             "instances": self.collect_instances,
+            "lambda_functions": self.collect_lambda_functions,
             "volumes": self.collect_volumes,
             "snapshots": self.collect_snapshots,
             "elbs": self.collect_elbs,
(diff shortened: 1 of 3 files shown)


lloesche:
@fernandocarletti thank you for the PR! Code looks great. 🥳 
One question and one comment I had while reading the code:

Question:
`role`, `runtime`, `code_size`, etc. are all typed `Optional`, but during collection the code says e.g.
```
lambda_function.runtime = function["Runtime"]
``` 
Is `runtime` really `Optional`? If yes then it should be `lambda_function.runtime = function.get("Runtime")`. Or, if you know that the boto3 API will always return a `Runtime` then the `runtime` attribute on the class shouldn't be `Optional`.

Also you could pass all those attributes directly to the constructor instead of setting them "later". 

So like
```
            lambda_function = AWSLambdaFunction(
                arn,
                tags,
                account=self.account,
                region=region,
                name=name,
                arn=arn,
                runtime=function["Runtime"],
                code_size=int(function["CodeSize"]),
                memory_size=function["MemorySize"],
                mtime=function["LastModified"],
                role=function["Role"],
                kms_key_arn=function.get("KmsKeyArn"),
              )

```

Again thank you for the PR! 👍

fernandocarletti:
They are not optional. TBH I had it in the back of my mind to revisit it but ended up forgetting about it. I'll make those adjustments.

I was trying to find a way to unit test those changes, but I could not find an example of that. Can you point me to any example, please?

lloesche:
@fernandocarletti re unit tests, skip them for now. We're currently rewriting the AWS collector. If you check the tests directory you'll see a bunch of unit tests for all kinds of boto3 methods. The old collector was not written in a way that would allow for easy mocking of the remote responses. Once merged we'll port your lambda collection.

lloesche:
@fernandocarletti do you feel this is ready to merge?

resotolib/resotolib/baseresources.py
```diff
@@ -854,6 +854,12 @@ class BaseBucket(BaseResource):
     kind: ClassVar[str] = "bucket"
 
 
+@define(eq=False, slots=False)
+class BaseServerlessFunction(BaseResource):
+    kind: ClassVar[str] = "serverless_function"
+    fingerprint: str = ""
```
aquamatthias, line 860: Just stumble upon this property, which is not filled in the lambda collector. Why is this?

plugins/aws/resoto_plugin_aws/accountcollector.py
```diff
@@ -1244,6 +1249,62 @@ def collect_instances(self, region: AWSRegion, graph: Graph) -> None:
...
+                    graph.add_edge(subnet, lambda_function)
+                    graph.add_edge(subnet, lambda_function, edge_type=EdgeType.delete)
+
+            for security_group_id in vpc_config.get("SecurityGroupIds", {}):
+                security_group = graph.search_first("id", security_group_id)
+                if security_group:
+                    graph.add_edge(security_group, lambda_function)
+                    graph.add_edge(security_group, lambda_function, edge_type=EdgeType.delete)
+
+            # TODO get ENIs
```
aquamatthias, line 1304: Is this a left over? Where would the ENIs come from?

fernandocarletti:
@lloesche I still have some changes to apply here. I have a weird type mismatch in one of the fields that I need to fix. I'm having some busy weeks, I could not get back to it, but I plan to finish It by the weekend.

If you guys are in a hurry, feel free to finish it :)

plugins/aws/resoto_plugin_aws/accountcollector.py
fernandocarletti: That was a note for me. Sadly it seems there's no clean way to get them. Although the IPs get assigned the only clear relation I could find in the API was the description of the ENI, which didn't seem like a scalable way to query it.

fernandocarletti:
@lloesche I just fixed some conflicts and fixed the issues mentioned in the review. It should be ready. Thanks for your patience!

fernandocarletti:
Actually, never mind. I just noticed it was superseded by #1022. I'm closing this one!

lloesche:
@fernandocarletti it'll be a bit before #1022 will be active. My best estimate: around 2 weeks. I'm fine with merging your code into the existing collector. Code looks good to me! 👍 

Like I mentioned, we're refactoring the entire AWS collector, but before we activate the new code a series of extensive tests need to be run in different user's real world environments, to make sure it's on par with the existing collector code and doesn't break anything. If you're interested in that, #1023 is the draft PR that has the new code activated.

The existing code is used in production environments so we're very careful about not disturbing those, even on `main`. Esp. the cleanup functionality is crucial and before we activate such a massive code update that essentially replaces the entire collector we'll do lots of manual tests for every single resource.

The new collector code has the advantage that all resources return 100% of the information that the AWS API returns. And also adding new AWS services has become much much easier. Plus full test coverage.

lloesche (approved):

PR #1013, opened 2022-07-14
[plugins/aws][feat] Add auth test before collect

lloesche:
# Description

- add auth test before collect
- do not send graph if no replace nodes found

This is in preparation of a `pre_flight()` method. Also abort collect runs with invalid graphs sooner and don't even attempt to send the graph to the core.

# Code of Conduct


aquamatthias (approved):
thanks

PR #1066, opened 2022-08-05
[plugins/aws][feat] Add support for SQS

anjafr:
# Description

# To-Dos

- Fixes #XXXX

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/resource/sqs.py
+++ b/plugins/aws/resoto_plugin_aws/resource/sqs.py
@@ -56,21 +56,30 @@ class AwsSqsQueue(AwsResource):
 
     @classmethod
     def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> None:
+        def add_attributes(queue: AwsSqsQueue) -> None:
+            queue_attributes = builder.client.call(
+                "sqs", "get-queue-attributes", "Attributes", QueueUrl=queue.sqs_queue_url, AttributeNames=["All"]
+            )
+            queue_attributes["QueueUrl"] = queue.sqs_queue_url
+            queue_attributes["QueueName"] = cast(str, queue.sqs_queue_url).rsplit("/", 1)[-1]
+            # instance = cls.from_api(js)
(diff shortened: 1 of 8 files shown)


plugins/aws/resoto_plugin_aws/resource/sqs.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @classmethod
+    def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> None:
+        def add_tags(queue: AwsSqsQueue) -> None:
+            tags = builder.client.list("sqs", "list-queue-tags", result_name="Tags", QueueUrl=[queue.sqs_queue_url])
+            if tags:
+                queue.tags = cast(Dict[str, Optional[str]], tags)
+
+        for queue_url in json:
+            queue_attributes = builder.client.call(
```
meln1k, line 65: Wdyt about not doing this call here, but rather using an approach similar to add_tags? this way we won't block the current thread for the http call.

anjafr: I don't get an approach like that to work . With only the `QueueUrl` from the first `list-queues` api call it's not possible to create an instance with `instance = cls.from_api(js)` 🤔

meln1k: I was thinking about something like:
```
for queue_url in json:
    builder.submit_work(add_instance, queue_url)
    ...
 ```
 
 and move the content of the existing loop into the `add_instance` method

File: plugins/aws/tools/model_gen.py [hunk 2]
```diff
@@ -616,6 +616,11 @@ def all_models() -> List[AwsModel]:
     "service-quotas": [
         # AwsResotoModel("list-service-quotas", "Quotas", "ServiceQuota", prefix="Quota", prop_prefix="quota_")
     ],
+    "sqs": [
+        AwsResotoModel(
```
meln1k, line 620: please comment it out :)

PR #1084, opened 2022-08-17
[resotocore][feat] Check the updated model not the update

aquamatthias:
# Description

The existing model is not checked for the model update.
The only check we will do: the final model needs to be sound and should not have any overlapping property paths with different types.

# To-Dos

# Code of Conduct


meln1k (approved):
LGTM

PR #1107, opened 2022-08-22
[aws][chore] Enable sandbox testing not maintained by git

anjafr:
# Description

Provides the `.sandbox/` directory, the contents of which are being ignore by git and thus can be used for experimental sandbox-y files that should not be tracked.

# Code of Conduct


.gitignore
```diff
@@ -32,3 +32,5 @@ resotocore/tools/Assets
 out/
 *.iml
 **/.hypothesis/
+plugins/aws/test/resources/*.ipynb
```
aquamatthias, line 35: I guess this is might be helpful not only in the context of AWS - maybe we use a directory on project root level?

plugins/aws/resoto_plugin_aws/aws_client.py
```diff
@@ -41,6 +41,9 @@ def __init__(
         self.role = role
         self.profile = profile
         self.region = region
+        self.AWS_ACCESS_KEY_ID = None
+        self.AWS_SECRET_ACCESS_KEY = None
+        self.AWS_SESSION_TOKEN = None
```
aquamatthias, line 46: I don't think this is a good idea. Is there a better way to achieve this without changing production code?

aquamatthias (changes requested):

plugins/aws/resoto_plugin_aws/aws_client.py
anjafr: @aquamatthias all of this is removed, credentials are now being used via the corresponding credentials file (or other mechanisms that I don't know about )

.gitignore
```diff
@@ -32,3 +32,6 @@ resotocore/tools/Assets
 out/
 *.iml
 **/.hypothesis/
+plugins/api_sandboxes/*.ipynb
+!plugins/api_sandboxes/*.example.ipynb
```
aquamatthias, line 36: Let's remove this line and the corresponding example file.
Create a `README.md` file that explains the reason for this folder, ala:

```
All files placed in this directory are ignored by git.
You can use this directory to put files and directories for sandbox testing.
```

.gitignore
```diff
@@ -32,3 +32,6 @@ resotocore/tools/Assets
 out/
 *.iml
 **/.hypothesis/
+plugins/api_sandboxes/*.ipynb
```
aquamatthias, line 35: Why plugins?
I would opt for `.sandbox/` - the dot for hiding this directory and everything in it (no matter which extension).

aquamatthias (commented):
Thanks!

To make this use case available without being prominent, I would opt for a dot directory on root level.

aquamatthias (approved):
Thanks!

TheCatLady (approved):

PR #1113, opened 2022-08-23
[plugins/aws][feat] Collect tags and enable tagging methods for Route53 Hosted Zones

anjafr:
# Description

Collect tags and enable tagging methods for Route53 Hosted Zones.

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/resource/route53.py
+++ b/plugins/aws/resoto_plugin_aws/resource/route53.py
@@ -1,6 +1,7 @@
 from typing import ClassVar, Dict, Optional, List, Type, cast
 
 from attrs import define, field
+from resoto_plugin_aws.aws_client import AwsClient
 
 from resoto_plugin_aws.resource.base import AwsResource, AwsApiSpec, GraphBuilder
 from resoto_plugin_aws.utils import ToDict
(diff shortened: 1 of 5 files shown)


anjafr:
@aquamatthias I'm debating whether it would make more sense do to the `rsplit("/", 1)[-1]` bit as soon as we map the Hosted Zone Id and then use it like that instead of rsplitting that string for every further call. 
The context is that what we get in the `list-hosted-zones`-response for the ID looks smth like `/hostedzones/abcdefg`. But the `ResourceId` or `Id` Parameter that is required for those other calls only ever expects the  `abcdefg` part of it. You get an error if you put the whole thing as argument. So maybe we should just remove the rest of it right away?

lloesche:
Is the whole `/hostedzone/` string ever used anywhere? If not, can't we just remove it from the start and just store the ID? Then we don't have to strip it for tag updates/retrieval.

anjafr:
> Is the whole `/hostedzone/` string ever used anywhere? If not, can't we just remove it from the start and just store the ID? Then we don't have to strip it for tag updates/retrieval.

see above comment 🤣

aquamatthias:
> see above comment 🤣

So it looks like we do not lose information, have proper ids, and make our life easier.
Let's cut the id in the mapping.

lloesche:
> > Is the whole `/hostedzone/` string ever used anywhere? If not, can't we just remove it from the start and just store the ID? Then we don't have to strip it for tag updates/retrieval.
> 
> see above comment 🤣

lol 😂 nvm I'm blind

anjafr:
> So it looks like we do not lose information, have proper ids, and make our life easier.
> Let's cut the id in the mapping.

@aquamatthias @lloesche Done

aquamatthias (approved):
thanks!

PR #1123, opened 2022-08-24
[plugins/aws][feat] Collect tags for S3 Buckets

anjafr:
# Description

Collect tags for S3 Buckets

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/resource/s3.py
+++ b/plugins/aws/resoto_plugin_aws/resource/s3.py
@@ -4,9 +4,9 @@ from attrs import define
 import botocore.exceptions
 from resoto_plugin_aws.resource.base import AwsResource, AwsApiSpec, GraphBuilder
 from resotolib.baseresources import BaseBucket, BaseAccount  # noqa: F401
-from resotolib.json_bender import Bender, S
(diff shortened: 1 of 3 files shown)


plugins/aws/resoto_plugin_aws/resource/s3.py
```diff
@@ -20,10 +20,18 @@ class AwsS3Bucket(AwsResource, BaseBucket):
 
     @classmethod
     def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> None:
+        def add_tags(bucket: AwsS3Bucket) -> None:
+            tags = builder.client.call(
```
aquamatthias, line 24:
```suggestion
            tags = builder.client.list(
```

aquamatthias (approved):
Thanks.

I realize there are a lot of `call` usages already. We should make it private and not use it.
Let's try to use `list` or `get` everywhere instead of `call`. 
Can be done in a separate PR.

plugins/aws/resoto_plugin_aws/resource/s3.py
anjafr: Noted!

PR #1124, opened 2022-08-24
[aws][fix] Handle S3 "NoSuchTagSet" Exception in all cases

anjafr:
# Description
resource tag collection now handles "NoSuchTagSet" exceptions as well

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/resource/s3.py
+++ b/plugins/aws/resoto_plugin_aws/resource/s3.py
@@ -1,12 +1,12 @@
-from typing import ClassVar, Dict, List, Type, Optional  # noqa: F401
+from typing import ClassVar, Dict, List, Type, Optional, cast  # noqa: F401
 
 from attrs import define
 import botocore.exceptions
 from resoto_plugin_aws.resource.base import AwsResource, AwsApiSpec, GraphBuilder
 from resotolib.baseresources import BaseBucket, BaseAccount  # noqa: F401
-from resotolib.json_bender import Bender, S, bend
+from resotolib.json_bender import Bender, S
 from resoto_plugin_aws.aws_client import AwsClient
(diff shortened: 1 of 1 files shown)


plugins/aws/resoto_plugin_aws/resource/s3.py
```diff
@@ -20,18 +20,16 @@ class AwsS3Bucket(AwsResource, BaseBucket):
...
     def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> None:
-        def add_tags(bucket: AwsS3Bucket) -> None:
-            tags = builder.client.call(
-                service="s3", action="get_bucket_tagging", result_name="TagSet", Bucket=bucket.name
-            )
+        def add_tags(bucket: AwsS3Bucket, client: AwsClient) -> None:
+            tags = bucket._get_tags(client)
             if tags:
-                bucket.tags = bend(ToDict(), tags)
+                bucket.tags = cast(Dict[str, Optional[str]], tags)
```
meln1k, line 26: WDYT about changing the return type of the _get_tags method instead of casting it to the right type?

lloesche (approved):

plugins/aws/resoto_plugin_aws/resource/s3.py
lloesche: Sure, current return type is `Dict[str, str]` which seems very compatible to `Dict[str, Optional[str]]`. That said, in AWS there's always a value, so `Optional[str]` seems wrong?!

anjafr: this would mean changing all the types for the follow up methods like `set_tags` too. And an optional value in `set_tags` specifically doesn't seem right 🤔 Either way this is the more minimal change

PR #1141, opened 2022-08-31
[resotocore][feat] Show commands in sections

aquamatthias:
# Description
image

# To-Dos

# Code of Conduct

### DIFF
--- a/resotocore/Makefile
+++ b/resotocore/Makefile
@@ -84,8 +84,8 @@ venv-pypy:
 
 setup: clean clean-env venv-pypy
 
-list-outdated: 
+list-outdated:
 	pip list --outdated
 
-install-latest: 
+install-latest:
 	pip list --outdated --format=freeze | grep -v '^\-e' | cut -d = -f 1  | xargs -n1 pip install -U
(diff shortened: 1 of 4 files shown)


scapecast:
This is much more intuitive. The "Misc" category is still the largest category though, which tells me we can further categorize. 

feels like there is a "Transform" or "manipulate" type of category in there? and an "output"?

WDYT of this?

Output
------
echo - Send the provided message to downstream.
env - Retrieve the environment and pass it to the output stream.
json - Parse ison and pass parsed obiects to the output stream.

Transform
------
chunk - Chunk incoming elements in batches.
flatten - Take incoming batches of elements and flattens them to a stream of single elements.
head - Return n first elements of the stream.
jg - Filter and process json.
tail - Return n last elements of the stream.
uniq - Remove all duplicated objects from the stream.
write - Writes the incoming stream of data to a file in the defined format.

move into -->  "Setup":
--------
help - Shows available commands, as well as help for any specific command.
sleep - Suspend execution for an interval of time.

lloesche (approved):

aquamatthias:
> This is much more intuitive. The "Misc" category is still the largest category though, which tells me we can further categorize.


I agree. I still think that more categories do not make it easier. Would start with this and wait for responses.

PR #1147, opened 2022-09-02
[plugins/aws] Add support for ApiGateway

anjafr:
# Description

New resources:

- `aws_api_gateway_resource`
- `aws_api_gateway_authorizer`
- `aws_api_gateway_stage`
- `aws_api_gateway_deployment`
- `aws_api_gateway_rest_api`

# To-Dos

# Code of Conduct


anjafr:
@aquamatthias I had to temporarily disable the edge from an `Authorizer` to a `LambdaFunction` because it would make circular imports. Do you see another way or do we want to drop this edge entirely? See [here](https://github.com/someengineering/resoto/blob/870c0f2c03d5aaa8502d2eec9cfb4b70c71f25a3/plugins/aws/resoto_plugin_aws/resource/apigateway.py#L211)

plugins/aws/resoto_plugin_aws/resource/apigateway.py
```diff
@@ -0,0 +1,447 @@
...
+            restApiId=self.api_link,
+            stageName=self.name,
+        )
+        return True
+
+
+@define(eq=False, slots=False)
+class AwsApiGatewayDeployment(AwsResource):
+    kind: ClassVar[str] = "aws_api_gateway_deployment"
+    reference_kinds: ClassVar[ModelReference] = {"successors": {"default": ["aws_api_gateway_stage"]}}
```
aquamatthias, line 301: There is no `connect_in_graph` method, but a `reference_kinds` is defined: the idea would be to have this variable everywhere you create a connection. This way it is possible to see reflect the edges created and it can be maintained/updated accodingly.

plugins/aws/resoto_plugin_aws/resource/apigateway.py
```diff
@@ -0,0 +1,447 @@
...
+
+    @classmethod
+    def called_apis(cls) -> List[AwsApiSpec]:
+        return [
+            cls.api_spec,
+            AwsApiSpec("apigateway", "get-deployments"),
+            AwsApiSpec("apigateway", "get-stages"),
+            AwsApiSpec("apigateway", "get-authorizers"),
+            AwsApiSpec("apigateway", "get-resources"),
+        ]
```
aquamatthias, line 359: let's try to have a section for properties and a section for methods.
mixing them up, makes it hard to read.

plugins/aws/resoto_plugin_aws/resource/apigateway.py
```diff
@@ -0,0 +1,447 @@
...
+
+    @classmethod
+    def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> None:
+        for js in json:
+            api_instance = cls.from_api(js)
+            region = builder.region
+            api_instance.arn = f"arn:{arn_partition(region)}:apigateway:{region.id}::/restapis/{api_instance.id}"
+            builder.add_node(api_instance, js)
+            for deployment in builder.client.list("apigateway", "get-deployments", "items", restApiId=api_instance.id):
+                deploy_instance = AwsApiGatewayDeployment.from_api(deployment)
```
aquamatthias, line 394: Consider to add a comment in AwsApiGatewayDeployment, that this resource is collected via AwsApiGatewayRestApi. Same for other resources collected here.

plugins/aws/resoto_plugin_aws/resource/apigateway.py
```diff
@@ -0,0 +1,447 @@
...
+    api_endpoint_configuration: Optional[AwsApiGatewayEndpointConfiguration] = field(default=None)
+    api_policy: Optional[str] = field(default=None)
+    api_disable_execute_api_endpoint: Optional[bool] = field(default=None)
+
+    @classmethod
+    def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> None:
+        for js in json:
+            api_instance = cls.from_api(js)
+            region = builder.region
+            api_instance.arn = f"arn:{arn_partition(region)}:apigateway:{region.id}::/restapis/{api_instance.id}"
```
aquamatthias, line 391: Mmmh - I have seen this kind of method in different places. Can we make it dry?

aquamatthias:
> @aquamatthias I had to (temporarily?) disable the edge from an `Authorizer` to a `LambdaFunction` because it would make circular imports. Do you see another way or do we want to drop this edge entirely? See [apigateway.py, line 211](https://github.com/someengineering/resoto/blob/870c0f2c03d5aaa8502d2eec9cfb4b70c71f25a3/plugins/aws/resoto_plugin_aws/resource/apigateway.py#L211)

One option to search a node is using the class, which requires an import.
You can avoid this filter, if you use `kind="other_node_kind"` as filter criteria instead of the clazz filter.

plugins/aws/resoto_plugin_aws/resource/apigateway.py
```diff
@@ -0,0 +1,447 @@
...
+
+        # the following edge to a lambda function would lead to circular import errors
+        # if self.authorizer_uri:
+        #     lambda_name = self.authorizer_uri.split(":")[-1].removesuffix("/invocations")
+        #     builder.dependant_node(
+        #         self,
+        #         clazz=AwsLambdaFunction,
+        #         name=lambda_name,
+        #     )
+        if self.authorizer_credentials:
```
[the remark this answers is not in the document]
anjafr: siehe [line 189](https://github.com/someengineering/resoto/blob/bed09842f5531bc41d6418205d7fd58bc295e95b/plugins/aws/resoto_plugin_aws/resource/apigateway.py#L189) ?

plugins/aws/resoto_plugin_aws/resource/apigateway.py
anjafr: Yeah I see. Trouble here is that the edge from Deployment to Stage is created in the RestApi model instance (line 408). I did not see how else to reflect this connection?

plugins/aws/resoto_plugin_aws/resource/apigateway.py
```diff
@@ -0,0 +1,447 @@
...
+    authorizer_auth_type: Optional[str] = field(default=None)
+    authorizer_uri: Optional[str] = field(default=None)
+    authorizer_credentials: Optional[str] = field(default=None)
+    authorizer_identity_source: Optional[str] = field(default=None)
+    authorizer_identity_validation_expression: Optional[str] = field(default=None)
+    authorizer_result_ttl_in_seconds: int = field(default=None)
+    api_link: str = field(default=None)
+
+    def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
+        # TODO add edge to Cognito User Pool when applicable (via self.authorizer_provider_arns)
```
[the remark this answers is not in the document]
anjafr: 😊

plugins/aws/resoto_plugin_aws/resource/apigateway.py
anjafr: Agreed. Ideally this would be a method on all `AwsResource` subclasses, no? Happy to tackle that but would prefer to do that in a separate PR.

aquamatthias: Something "aws global" - AwsResource sounds valid. Fine to move it out of this PR.

plugins/aws/resoto_plugin_aws/resource/apigateway.py
aquamatthias: Gotcha. Then I think it is fine as it is. Maybe add 2 comments: one at `reference_kinds` saying where it is defined and one in RestApi, so people do not forget to update.

aquamatthias (approved):
Thanks!

PR #1158, opened 2022-09-08
[docker][fix] Define container names in Compose YAML

TheCatLady:
# Description

Define `container_name` for each service.

# To-Dos

- [x] Document new or updated functionality (someengineering/resoto.com#350)

# Code of Conduct

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -29,7 +29,7 @@ services:
       - --web.console.libraries=/usr/share/prometheus/console_libraries
       - --web.console.templates=/usr/share/prometheus/consoles
   resotocore:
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -76,27 +81,8 @@ services:
     restart: always
     stop_grace_period: 2m
   resotoshell:
-    # This container is not started by default. To start a shell use this command:
```
aquamatthias, line 79: I generally agree to remove this option -let's not remove it right away.
there might be people that still use it this way.

lloesche: This is going into `main` 3.0 though, so I say let's break it now.

lloesche (approved):

PR #1170, opened 2022-09-14
[aws][fix] Map ctime and mtime for aws_sqs_queue correctly

anjafr:
# Description

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/resource/sqs.py
+++ b/plugins/aws/resoto_plugin_aws/resource/sqs.py
@@ -21,8 +21,8 @@ class AwsSqsQueue(AwsResource):
     mapping: ClassVar[Dict[str, Bender]] = {
         "id": S("QueueName"),
         "name": S("QueueName"),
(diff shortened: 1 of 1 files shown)


plugins/aws/resoto_plugin_aws/resource/sqs.py
```diff
@@ -20,6 +21,8 @@ class AwsSqsQueue(AwsResource):
     mapping: ClassVar[Dict[str, Bender]] = {
         "id": S("QueueName"),
         "name": S("QueueName"),
+        "ctime": S("CreatedTimestamp") >> F(lambda x: utc_str(datetime.fromtimestamp(x))),
```
aquamatthias, line 24: mmhh I would assume that the timestamp is based on UTC (guessing) - but it is interpreted with local time (fromtimestamp). maybe we should interpret this timestamp in UTC as well?

aquamatthias (approved):
Thanks!

PR #1171, opened 2022-09-15
[aws] Add methods to handle missing ARNs/IDs

anjafr:
# Description

Introduces
- a method on `AwsResource` to construct the ARN of a resource instance.
- a static method on `AwsResource` to extract the ID from a given ARN

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/resource/base.py
+++ b/plugins/aws/resoto_plugin_aws/resource/base.py
@@ -16,6 +16,7 @@ from boto3.exceptions import Boto3Error
 from resoto_plugin_aws.config import AwsConfig
 from resoto_plugin_aws.aws_client import AwsClient
 from resoto_plugin_aws.resource.pricing import AwsPricingPrice
+from resoto_plugin_aws.utils import arn_partition
 from resotolib.baseresources import (
     BaseResource,
     EdgeType,
@@ -65,6 +66,24 @@ class AwsResource(BaseResource, ABC):
 
     # The AWS specific identifier of the resource. Not available for all resources.
     arn: Optional[str] = None
+
+    def set_arn(
+        self,
+        builder: GraphBuilder,
+        region: Optional[str] = None,
+        service: Optional[str] = None,
+        account: Optional[str] = None,
+        resource: str = "",
+    ) -> None:
+        aws_region = builder.region
+        partition = arn_partition(aws_region)
+        if region is None:
+            region = aws_region.id
+        if service is None and self.api_spec:
+            service = self.api_spec.service
+        if account is None:
+            account = builder.account.id
+        self.arn = f"arn:{partition}:{service}:{region}:{account}:{resource}"
 
     # TODO: implement me
     def update_resource_tag(self, client: AwsClient, key: str, value: str) -> bool:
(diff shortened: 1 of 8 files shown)


plugins/aws/resoto_plugin_aws/resource/base.py
```diff
@@ -119,6 +120,24 @@ def to_json(self) -> Json:
...
+        builder: GraphBuilder,
+        region: Optional[str] = None,
+        service: Optional[str] = None,
+        account: Optional[str] = None,
+        resource: str = "",
+    ) -> None:
+        aws_region = builder.region
+        partition = arn_partition(aws_region)
+        if region is None:
+            region = aws_region.id
```
aquamatthias, line 134: wouldn't we need to handle the global region different (like empty)?

plugins/aws/resoto_plugin_aws/resource/base.py
```diff
@@ -119,6 +120,24 @@ def to_json(self) -> Json:
             ),
         )
 
+    def set_arn(
+        self,
+        builder: GraphBuilder,
+        region: Optional[str] = None,
+        service: Optional[str] = None,
+        account: Optional[str] = None,
+        resource: str = "",
```
aquamatthias, line 129: According to https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-paths there is also an optional resource type. Consider to add.

aquamatthias (commented):
I wonder if it would be possible to create the arn for any AWS resource?

plugins/aws/resoto_plugin_aws/resource/base.py
anjafr: For everything after the account it, there's a lot of variation. Can be either of
```
arn:partition:service:region:account-id:resource-id
arn:partition:service:region:account-id:resource-type/resource-id
arn:partition:service:region:account-id:resource-type:resource-id
```
The method as written is bundling this under `resource`, because you'd need a whole lot of context otherwise. See for example:
```py
deploy_instance.set_arn(
    builder=builder,
    account="",
    resource=f"/restapis/{api_instance.id}/deployments/{deploy_instance.id}",
)
```

plugins/aws/resoto_plugin_aws/resource/base.py
anjafr: In that case you put `region=""`, same for `account` which isn't always filled. See [this arn for apigateway for example](https://github.com/someengineering/resoto/blob/3ed3f3be24c4b212f028b4f5c47b4dc401438afd/plugins/aws/test/resources/apigateway_test.py#L13)

anjafr:
> I wonder if it would be possible to create the arn for any AWS resource?

Technically: ... maybe - but practically it would require an enormous amount of contextual information that I don't think is feasible to gather here

anjafr:
> > I wonder if it would be possible to create the arn for any AWS resource?
> 
> Technically: ... maybe - but practically it would require an enormous amount of contextual information that I don't think is feasible to gather here

@aquamatthias Did you have a chance to mull this over?

aquamatthias:
So the TL;DR of our discussion: it is impossible to generate the ARN based on the information we already have. We need to handle this case manually in every service that does not report the ARN.

aquamatthias (approved):
Thanks!

PR #1181, opened 2022-09-19
[plugins/aws][fix] Do not collect the default AwsDataCatalog

meln1k:
# Description

Excluded the default data catalog from collection

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/resource/athena.py
+++ b/plugins/aws/resoto_plugin_aws/resource/athena.py
@@ -210,7 +210,7 @@ class AwsAthenaDataCatalog(AwsResource):
                 data_catalog.tags = bend(S("Tags", default=[]) >> ToDict(), tags[0])
 
         for js in json:
-            if (name := js.get("CatalogName")) is not None and isinstance(name, str):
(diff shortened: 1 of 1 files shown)


plugins/aws/resoto_plugin_aws/resource/athena.py
```diff
@@ -210,7 +210,7 @@ def add_tags(data_catalog: AwsAthenaDataCatalog) -> None:
                 data_catalog.tags = bend(S("Tags", default=[]) >> ToDict(), tags[0])
 
         for js in json:
-            if (name := js.get("CatalogName")) is not None and isinstance(name, str):
+            if (name := js.get("CatalogName")) is not None and isinstance(name, str) and name != "AwsDataCatalog":
```
aquamatthias, line 213: Can you add a comment that explains why this catalog is sorted out?
Should we sort out all catalogs of type `GLUE`?

meln1k: > Should we sort out all catalogs of type GLUE?

I don't think so.

aquamatthias (approved):
thanks!

PR #1198, opened 2022-09-28
[resotolib][feat] Add more log levels

lloesche:
# Description

Add more log levels. Used by AWS plugin for pagination calls.

# Code of Conduct

### DIFF
--- a/resotolib/resotolib/logger.py
+++ b/resotolib/resotolib/logger.py
@@ -3,6 +3,7 @@ from attrs import define, field
 from typing import ClassVar, Optional, Dict, Mapping
 import sys
 import os
+import logging
 from logging import (
     basicConfig,
     getLogger,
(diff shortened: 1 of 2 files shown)


resotolib/resotolib/logger.py
```diff
@@ -15,9 +16,16 @@
...
 )
 
+
 from resotolib.args import ArgumentParser
 from resotolib.types import Json
 
+DEBUG2 = DEBUG - 1
+DEBUG3 = DEBUG - 2
+DEBUG4 = DEBUG - 3
+DEBUG5 = DEBUG - 4
```
aquamatthias, line 26: no strong objection - isn't introducing trace sufficient?
if you want to filter/grep debug, you would also get debug2 ...

lloesche: I've often wished for much finer debug output levels, esp. when talking to remote APIs. Trace to me is what I select as the user, and the different debug levels is what I'd use as the log producer. E.g. debug4 for raw packet data, debug3 could be the parsing information of those packets, debug2 could be decisions taken based on the parsed data. But the user would just turn on trace logging to see all of them and then within the log output have sort of categories of log information where they can jump fron one raw packet to the next for example.

aquamatthias: Personally, I think it is too complicated to judge if a message is `debug2` or `debug4` (while both are actually not visible when debug is enabled..).

aquamatthias (approved):

PR #1219, opened 2022-10-12
[resoto][feat] CDK construct

meln1k:
# Description

Added a CDK construct for a one command resoto deployment to AWS.

# To-Dos

- [ ] Store container data on a different volume 

# Code of Conduct

### DIFF
--- a/cdk/lib/resoto-stack.ts
+++ b/cdk/lib/resoto-stack.ts
@@ -2,6 +2,7 @@ import * as cdk from 'aws-cdk-lib';
 import { Construct } from 'constructs';
 import * as ec2 from 'aws-cdk-lib/aws-ec2';
 import * as iam from 'aws-cdk-lib/aws-iam';
+import { readFileSync } from 'fs';
 
 export class ResotoStack extends cdk.Stack {
   constructor(scope: Construct, id: string, props?: cdk.StackProps) {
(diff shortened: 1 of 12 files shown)


cdk/lib/resoto-stack.ts
```diff
@@ -0,0 +1,94 @@
...
+    super(scope, id, props);
+
+    const vpc = new ec2.Vpc(this, 'resoto-vpc', {
+      maxAzs: 1,
+      subnetConfiguration: [{
+          name: 'public', 
+          cidrMask: 24,
+          subnetType: ec2.SubnetType.PUBLIC,
+        }]
+    });
```
aquamatthias, line 17: I am wondering if we should support deploying it into an existing vpc?

cdk/lib/resoto-stack.ts
```diff
@@ -0,0 +1,94 @@
...
+      securityGroup: resotocoreSG,
+      instanceType: ec2.InstanceType.of(
+        ec2.InstanceClass.T3,
+        ec2.InstanceSize.LARGE
+      ),
+      machineImage: image,
+      blockDevices: [
+        {
+          deviceName: '/dev/xvda', // make the root volume bigger
+          volume: ec2.BlockDeviceVolume.ebs(50),
```
aquamatthias, line 78: I guess the size should be configurable?

cdk/lib/resoto-stack.ts
```diff
@@ -0,0 +1,94 @@
...
+        },
+      ],
+      keyName: 'nm-resoto-ec2-keypair',
```
aquamatthias, line 81: Where is this keypair coming from?

cdk/lib/resoto-stack.ts
```diff
@@ -0,0 +1,94 @@
...
+      'sudo mv docker-compose-$(uname -s)-$(uname -m) /usr/local/bin/docker-compose',
+      'sudo chmod -v +x /usr/local/bin/docker-compose',
+      'sudo systemctl enable docker.service',
+      'sudo systemctl start docker.service',
+      // resoto setup
+      'mkdir -p resoto/dockerV2',
+      'cd resoto',
+      'curl -o docker-compose.yaml https://raw.githubusercontent.com/someengineering/resoto/2.4.3/docker-compose.yaml',
+      'curl -o dockerV2/prometheus.yml https://raw.githubusercontent.com/someengineering/resoto/2.4.3/dockerV2/prometheus.yml',
+      'docker-compose up -d',
```
aquamatthias, line 59: This means it is docker version 1 - is it possible to user docker version 2?

aquamatthias (commented):
Great start - thanks.
I think this content should be placed into a separate repository.

lloesche:
My two cents: I'm very much against all of this single point of failure deployment model. Would MUCH more prefer deployment on EKS where failures are automatically resolved and data is persisted onto dedicated EBS volumes. If somebody can't use K8S they're likely the wrong user for Resoto anyways.

cdk/lib/resoto-stack.ts
meln1k: that's not straightforward using the amazon image.

cdk/lib/resoto-stack.ts
meln1k: needs to be configurable

cdk/lib/resoto-stack.ts
meln1k: yep

meln1k:
@lloesche @scapecast if k8s is what you wish, then so be it :) I will update the CDK construct to create a new EKS cluster in case our users do not have one yet, and then deploy a helm chart on it.

aquamatthias:
Moved to a separate repository: https://github.com/someengineering/resoto-cdk

Closing this PR.

PR #1221, opened 2022-10-13
[resotocore][feat] Add /system/data endpoint that returns SystemData for the UI

lloesche:
# Description

Add /system/data endpoint that returns SystemData for the UI

# To-Dos

Currently the web/api_tests.py checks the `/system/*` endpoints by utilizing resotoclient functions.
Resotoclient in turn uses these `/system/` endoints to provide `ready()` and `ping()`.
I would merge this PR to provide the `/system/data/` endpoint, then extend `resotoclient` to provide the `data()` function. And then update the resotocore tests to call `data()` and check that it contains `system_id`.

# Code of Conduct

### DIFF
--- a/resotocore/resotocore/web/api.py
+++ b/resotocore/resotocore/web/api.py
@@ -46,6 +46,7 @@ from resotocore.ids import TaskId, ConfigId, NodeId
 from resotocore.config import ConfigHandler, ConfigValidation, ConfigEntity
 from resotocore.console_renderer import ConsoleColorSystem, ConsoleRenderer
 from resotocore.core_config import CoreConfig
+from resotocore.db import SystemData
 from resotocore.db.db_access import DbAccess
 from resotocore.db.graphdb import GraphDB
 from resotocore.db.model import QueryModel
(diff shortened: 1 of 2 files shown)


resotocore/resotocore/web/api.py
```diff
@@ -226,6 +229,7 @@ def __add_routes(self, prefix: str) -> None:
                 # system operations
                 web.get(prefix + "/system/ping", self.ping),
                 web.get(prefix + "/system/ready", self.ready),
+                web.get(prefix + "/system/data", self.data),
```
aquamatthias, line 232: In the CLI we call our system into. Consider to rename the endpoint to info.

aquamatthias (commented):
Please also add Api docs for this endpoint

resotocore/resotocore/web/api.py
lloesche: `system info` is not the same as `/system/data`. info is version and memory info.
`/system/data` returns
```
{"created_at": "2022-01-22T12:12:22Z", "db_version": 1, "system_id": "8d7ff7ce-7b7c-11ec-91f2-1ad5c904c3a3"}
```

Pablo uses the system_id for Posthog in the UI. Although Raffa and I just had a talk whether it wouldn't make more sense for the Core to do all Posthog communication and the UI to send it's data to the core. Otherwise UI has to know about core settings like do not track and needs to get access to the core's `RESOTOCORE_ANALYTICS_OPT_OUT` environment variable.

lloesche: Maybe a proxy endpoint sort of like we have it for Prometheus makes more sense and then the core can decide whether to forward the data to posthog or to silently drop it. That way the UI does not have to be concerned with user analytics settings.

PR #1230, opened 2022-10-20
[resotolib][fix] return closest parent class

lloesche:
# Description

Returns the closest parent class instead of first best.
Also fixes aws client to use region id instead of name.

# Code of Conduct


meln1k (dismissed):
Nice! Thanks!

PR #1233, opened 2022-10-21
[resotolib][fix] When searching for parent nodes only check them once

lloesche:
# Description

There could be multiple paths to a node which would currently result in them being checked multiple times.

# Code of Conduct

### DIFF
--- a/resotolib/resotolib/graph/__init__.py
+++ b/resotolib/resotolib/graph/__init__.py
@@ -330,12 +330,15 @@ class Graph(networkx.MultiDiGraph):
         This is being used to search up the graph and e.g. find the account that the
         graph node is a member of.
         """
-        queue = deque()
-        queue.extend(self.predecessors(node))
+        queue = deque([self.predecessors(node)])
+        already_checked = set()
         while queue:
             current = queue.popleft()
+            if current in already_checked:
+                continue
             if isinstance(current, cls):
                 return current
+            already_checked.add(current)
             queue.extend(self.predecessors(current))
         return None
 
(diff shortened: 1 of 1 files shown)


resotolib/resotolib/graph/__init__.py
```diff
@@ -330,12 +330,15 @@ def search_first_parent_class(self, node, cls):
...
         graph node is a member of.
         """
-        queue = deque()
-        queue.extend(self.predecessors(node))
+        queue = deque(self.predecessors(node))
+        already_checked = set()
         while queue:
             current = queue.popleft()
+            if current in already_checked:
+                continue
```
meln1k, line 338:
```suggestion
```

resotolib/resotolib/graph/__init__.py
```diff
@@ -330,12 +330,15 @@ def search_first_parent_class(self, node, cls):
         This is being used to search up the graph and e.g. find the account that the
         graph node is a member of.
         """
-        queue = deque()
-        queue.extend(self.predecessors(node))
+        queue = deque(self.predecessors(node))
+        already_checked = set()
```
meln1k, line 334:
```suggestion
        already_checked = set(self.predecessors(node))
```

resotolib/resotolib/graph/__init__.py
```diff
@@ -330,13 +330,16 @@ def search_first_parent_class(self, node, cls):
...
```
meln1k, line 334: WDYT about this? Our implicit contract here is that add the new nodes and mark them as visited in a lockstep, and we want to do the same in the initialization step too.

```suggestion
        queue = deque(set(self.predecessors(node)))
        already_checked = set(self.predecessors(node))
```

meln1k (dismissed):
LGTM!

resotolib/resotolib/graph/__init__.py
meln1k, line 334:
```suggestion
        already_checked = set(self.predecessors(node))
```

meln1k (approved):

PR #1269, opened 2022-11-14
[plugins/aws][fix] Exception handling for individual resource kinds and rename of progress prefix

lloesche:
# Description

Update of the AWS collector:
- Handle exceptions on a kind level instead of aborting the entire region when an individual kind is not allowed to be collected.
- Update progress to use account name & id instead of just the id

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/__init__.py
+++ b/plugins/aws/resoto_plugin_aws/__init__.py
@@ -351,23 +351,23 @@ def current_account_id(profile: Optional[str] = None) -> str:
     return session.client("sts").get_caller_identity().get("Account")  # type: ignore
 
 
-def set_account_name(account: AwsAccount) -> None:
-    try:
-        account_aliases = (
-            aws_session(account.id, account.role, account.profile)
-            .client("iam")
-            .list_account_aliases()
-            .get("AccountAliases", [])
-        )
-        if len(account_aliases) > 0:
-            account.name = account_aliases[0]
-    except Exception:
-        pass
-
-
 def set_account_names(accounts: List[AwsAccount]) -> None:
+    def set_account_name(account: AwsAccount) -> None:
+        try:
+            account_aliases = (
+                aws_session(account.id, account.role, account.profile)
+                .client("iam")
+                .list_account_aliases()
+                .get("AccountAliases", [])
+            )
+            if len(account_aliases) > 0:
+                account.name = account_aliases[0]
+        except Exception:
+            pass
+
     if len(accounts) == 0:
         return
+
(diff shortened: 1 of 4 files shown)


plugins/aws/resoto_plugin_aws/__init__.py
```diff
@@ -349,6 +351,28 @@ def current_account_id(profile: Optional[str] = None) -> str:
...
+        if len(account_aliases) > 0:
+            account.name = account_aliases[0]
+    except Exception:
+        pass
+
+
+def set_account_names(accounts: List[AwsAccount]) -> None:
+    if len(accounts) == 0:
+        return
+    max_workers = len(accounts) if len(accounts) < Config.aws.account_pool_size else Config.aws.account_pool_size
```
aquamatthias, line 371: I guess this can be simplified to `Config.aws.account_pool_size` ?  if the limit is not reached, no new threads are created.

plugins/aws/resoto_plugin_aws/__init__.py
```diff
@@ -349,6 +351,28 @@ def current_account_id(profile: Optional[str] = None) -> str:
     return session.client("sts").get_caller_identity().get("Account")  # type: ignore
 
 
+def set_account_name(account: AwsAccount) -> None:
```
aquamatthias, line 354: Consider to move this function inside `set_account_names`

aquamatthias (approved):
Thanks!

plugins/aws/resoto_plugin_aws/__init__.py
lloesche: If you set max_workers to 50 and use a ProcessPool you can see 50 processes running. They are created upon creation of the pool, not as tasks are submitted to the pool. Seeing this behaviour in my local `ps` process list was what originally prompted me to add this optimization.

plugins/aws/resoto_plugin_aws/__init__.py
lloesche: Good point! Will do.

plugins/aws/resoto_plugin_aws/__init__.py
aquamatthias:
```python
with ThreadPoolExecutor(max_workers=128):
    print("active threads: ", threading.active_count())

with ProcessPoolExecutor(max_workers=128):
    print("active processes: ", len(psutil.Process().children(recursive=True)))

--outputs:
active threads:  1
active processes:  1
```

Docs say this:
> Changed in version 3.8: 
> ThreadPoolExecutor now reuses idle worker threads before starting max_workers worker threads too.

I guess it's OK to have an additional safeguard..

aquamatthias (approved):
Thanks!

PR #1281, opened 2022-11-15
[docker][fix] Optimize swap space

lloesche:
# Description

Optimize swap space.
The Cryptography ARM64 build requires more than the 10G swap we had previously. With 20G it went through but then failed with an out of disk space error later on. This splits the swap space across two different mount points.

# Code of Conduct


aquamatthias (approved):
nifty!

PR #1320, opened 2022-11-29
[resoto][fix] Allow worker and metrics to start webserver without TLS

lloesche:
# Description

Allow worker and metrics to start webserver without TLS
Also fixes an exception that can happen during config related reloads when there are no open files.

# Code of Conduct


aquamatthias (approved):
thanks!

PR #1355, opened 2022-12-21
[resotocore][feat] Allow to search nested complex objects inside arrays

aquamatthias:
# Description

This PR improves the capability of the Search DSL in Resoto. It allows to express patterns in objects inside arrays.
A simple predicate can look into arrays. Using the `[*]` it is possible to match any object inside an array.
If we have more than one predicate that we want to match inside an object inside an array, we can not simply use the `[*]` operator twice for each predicate, since each predicate matches on its own - there is no way to "bundle" 2 or more predicates. This PR introduces a context term, that allows exactly this.

Example using  an AWS security gateway (excerpt):
```json
{
  "id": "sg-123",
  "tags": {},
  "name": "redirus-master",
  "description": "redirus Master Security Group",
  "group_ip_permissions": [
    {
      "from_port": 80,
      "to_port": 80,
      "ip_protocol": "tcp",
      "ip_ranges": [
        { "cidr_ip": "0.0.0.0/0" }
      ]
    },
    {
      "from_port": 22,
      "to_port": 22,
      "ip_protocol": "tcp",
      "ip_ranges": [
        { "cidr_ip": "3.3.175.2/32" },
        { "cidr_ip": "4.4.119.97/32", "description": "Office access" }
      ]
    }
  ]
}
```

If we want to select gateways that allow ssh access (port 22) from everywhere, we need to select:
- filter for kind aws_security_gateway
- check for all group permissions that match: from_port >= 22, to_port <= 22 and at least one cidr that contains 0.0.0.0.

We could write a search that filters each single criteria:

```python
search is(aws_ec2_security_group) and group_ip_permissions[*].from_port>=22 and group_ip_permissions[*].to_port<=22 and group_ip_permissions[*].ip_ranges[*].cidr_ip~"0.0.0.0"
```

This query would eventually match more items than expected, since every filter criteria has to be fulfilled in at least one of the group permissions (e.g. there is one permission with from_port==to_port==22 but no cidr~"0.0.0.0" and another permission with cidr~"0.0.0.0" but with a different port).
What we actually wanted to express: all of the criteria should match in any of the nested complex objects.

The change made in this PR allows defining the context of the filter criteria, so it becomes clear, which filter criteria belong together and looks like this:

```python
search is(aws_ec2_security_group) and group_ip_permissions[*].{from_port>=22, to_port<=22, ip_ranges[*].cidr_ip~"0.0.0.0"} 
```

The term `group_ip_permission[*].{ xxxx }` says : look into the property group_ip_permission which is an array. Any element that matches the filters in { xxxx } will do. Instead of the [*] we could also define a number, which then would be interpreted as index of the array.

Every single group_ip_permission will be matched with the filters: `from_port>=22`, `to_port<=22`, `ip_ranges[*].cidr_ip~"0.0.0.0"`.

It would even be possible to nest context criterions. Example

```python
search is(aws_ec2_security_group) and group_ip_permissions[*].{from_port>=22, to_port<=22, ip_ranges[*].{ cidr_ip~"0.0.0.0", description!="Office access" }} 
```

# To-Dos

# Code of Conduct

### DIFF
--- a/resotocore/resotocore/cli/command.py
+++ b/resotocore/resotocore/cli/command.py
@@ -2346,7 +2346,7 @@ class ListCommand(CLICommand, OutputTransformer):
             # add all predicates the user has queried
             if ctx.query:
                 # add all predicates the user has queried
-                predicate_names = (p.name for p in ctx.query.predicates)
+                predicate_names = (p.name for p in ctx.query.visible_predicates)
                 # add sort keys of the last part the user has defined
                 sort_names = (s.name for s in ctx.query.current_part.sort)
                 for name in chain(predicate_names, sort_names):
--- a/resotocore/resotocore/db/arango_query.py
+++ b/resotocore/resotocore/db/arango_query.py
@@ -37,6 +37,7 @@ from resotocore.query.model import (
     SortOrder,
     FulltextTerm,
     Limit,
+    ContextTerm,
 )
 from resotocore.util import first, set_value_in_path, exist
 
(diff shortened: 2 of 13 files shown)


lloesche (approved):
Yeay 🥳 
This is soooo nice!

PR #1381, opened 2023-01-26
[docker][fix] set `provenance: false` on docker/build-push-action

TheCatLady:
# Description

Temporary workaround for docker/buildx#1533

# Code of Conduct

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -221,6 +221,7 @@ jobs:
             TESTS=false
           tags: ${{ steps.basemeta.outputs.tags }}
           labels: ${{ steps.basemeta.outputs.labels }}
+          provenance: false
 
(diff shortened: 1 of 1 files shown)


aquamatthias (dismissed):
Do we have people that are using resoto images on lambda?
Nevertheless - the change is fine - thanks. 
Please add a comment to the provenance=false lines (maybe even with a link to this issue),
so we know why it was disabled.
Thanks!

TheCatLady:
> Do we have people that are using resoto images on lambda?

Unsure, but this is a breaking change that doesn't only affect Lambda. https://github.com/docker/buildx/issues/1533#issuecomment-1404708819

> Please add a comment to the provenance=false lines (maybe even with a link to this issue), so we know why it was disabled. Thanks!

Will do. 👍🏻

aquamatthias (approved):

PR #1442, opened 2023-02-14
External configuration overrides

meln1k:
# Description

Added support for external configuration overrides. See https://docs.google.com/document/d/1ZWffnTnTPjUWi87y9fpIOXMtaZoJ_kOLMBdrsg99erM/edit# for more info.

# To-Dos

# Code of Conduct


resotocore/resotocore/core_config.py
```diff
@@ -622,13 +623,53 @@ def parse_config(args: Namespace, core_config: Json, command_templates: Optional
...
+    # json with all merged overrides
+    all_file_overrides: Json = {}
+    # merge all provided overrides into a single object, preferring the values from the last override
+    for path in args.config_override_path:
+        with path.open() as f:
+            raw_yaml = yaml.safe_load(f)
+            merged = merge_configs(all_file_overrides, raw_yaml)
+            all_file_overrides = merged
+
+    # replaces all values wrapped in $() with the environmental valiable
```
aquamatthias, line 644:
```suggestion
    # replaces all values wrapped in $() with the environmental variable
```

resotocore/tests/resotocore/core_config_test.py
```diff
@@ -92,6 +92,46 @@ def test_validate() -> None:
...
+        """,
+            encoding="utf-8",
+        )
+
+        # config that overrides the default config and containes an env var which can't be resolved
+        other_conf = Path(tmp, "config.yml")
+        other_conf.write_text(
+            """
+api:
+    web_port: 1337
```
aquamatthias, line 116: Please create a test where you replace an int (like the port) via env vars.
I wonder if the type checker will complain and needs special handling?

resotoworker/resotoworker/__main__.py
```diff
@@ -92,6 +94,28 @@ def main() -> None:
...
+            return [replace_env_vars(v) for v in elem]
+        elif isinstance(elem, str):
+            str_value = elem
+            for match in re.finditer(r"\$\((\w+)\)", elem):
+                if env_var_found := os.environ.get(match.group(1)):
+                    str_value = str_value.replace(match.group(0), env_var_found)
+
+            return str_value
+        else:
+            return elem
```
aquamatthias, line 112: Shouldn't this code live in resotolib?
I assume it is also relevant for resotometrics etc.
Also resotocore seems to use exactly the same functionality?

resotolib/resotolib/config.py
```diff
@@ -143,6 +146,10 @@ def load_config(self, reload: bool = False) -> None:
                 log.debug("Starting config event listener")
                 self._ce.start()
 
+    def add_config_load_hook(self, hook: Callable[[Json], Json]) -> None:
+        """Add a hook to be called after the config is loaded."""
+        self._config_load_hooks.append(hook)
+
```
aquamatthias, line 152: When would it make sense to have a configuration that does not replace env vars?
Maybe we should simply do this here automatically?

resotocore/tests/resotocore/cli/command_test.py
```diff
@@ -705,6 +705,17 @@ async def check_file_is_yaml(res: Stream) -> None:
     # show the entry - should be the same as the created one
     show_updated_result = await cli.execute_cli_command("configs show test_config", stream.list)
     assert show_updated_result[0][0] == update_doc
+    # write a env var substitution to the config
+    env_var_update = "foo: $(FOO)\n"
+    with open(config_file, "w") as file:
+        file.write(env_var_update)
+    ctx = CLIContext(uploaded_files={"config.yaml": config_file})
+    update_result = await cli.execute_cli_command(f"configs update test_config {config_file}", stream.list, ctx)
```
aquamatthias, line 713: You could do the same without writing to a file via:
```
configs set test_config foo="$(FOO)"
```

resotocore/resotocore/core_config.py
```diff
@@ -622,13 +623,53 @@ def parse_config(args: Namespace, core_config: Json, command_templates: Optional
...
+    def replace_env_vars(elem: JsonElement) -> JsonElement:
```
aquamatthias, line 645: See earlier comment - it looks like the same function is used elsewhere.

aquamatthias (commented):

lloesche (changes requested):
The first requirement from the design doc is not met currently:
```
The core needs to (implicitly) know about any overrides by being the only place where overrides/merges happen, so that it can show them in the UI/CLI when editing the config. As opposed to the current situation where overrides happen on the component (e.g. worker) without the core's knowledge.
```

There needs to be a mechanism by which the core knows which parts of the config have been merged from external sources and able to communicate this to the resh/UI. So that the user doesn't wonder why their config changes they do in the UI aren't applied on the system.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+ENV PYTHONUNBUFFERED 1
+
+# [Optional] If your requirements rarely change, uncomment this section to add them to the image.
+# COPY requirements.txt /tmp/pip-tmp/
+# RUN pip3 --disable-pip-version-check --no-cache-dir install -r /tmp/pip-tmp/requirements.txt \
+#    && rm -rf /tmp/pip-tmp
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
```
aquamatthias, line 12: can we remove this?

resotolib/resotolib/utils.py
```diff
@@ -514,3 +514,19 @@ def error(self, message: str) -> NoReturn:
     def exit(self, status: int = 0, message: Optional[str] = None) -> NoReturn:
         msg = message if message else "unknown"
         raise AttributeError(f"Could not parse arguments: {msg}")
+
+
+def replace_env_vars(elem: JsonElement) -> JsonElement:
```
aquamatthias, line 519: I would opt to pass the env vars as dict explicitly. A) easier to test B) we have control over the env vars that get passed. Please add a test.

resotolib/resotolib/config.py
```diff
@@ -77,6 +78,7 @@ def __init__(
             message_processor=self.on_config_event,
             tls_data=tls_data,
         )
+        self._config_load_hooks: List[Callable[[Json], Json]] = []
```
aquamatthias, line 81: can be removed?

resotocore/resotocore/dependencies.py
```diff
@@ -184,6 +185,19 @@ def key_value(kv: str) -> Tuple[str, JsonElement]:
...
+        nargs="+",
+        type=Path,
+        dest="config_override_path",
+        default=[],
+        help="Override configuration parameters via a YAML file. "
+        "The existing configuration will be patched with the provided values. "
+        "Note: this argument allows multiple overrides separated by space, in this case the "
+        "resulting configuration will be the merge of all the provided files, in the order they are provided."
+        "The same section can be overridden multiple times, in this case the last override will be used."
+        "Example: --override-file /path/to/your/config.yaml /path/to/your/other/config.yaml",
```
aquamatthias, line 199:
```suggestion
        "Example: --override-path /path/to/your/config.yaml /path/to/your/other/config.yaml",
```

aquamatthias: Also wonder: can we supply a directory here as well?

aquamatthias (changes requested):
Thanks.

I realized 2 issues:

1)
I tested this state with follwing override:
```
resotoworker:
  pool_size: $(POOLSIZE)
```

The configuration could not be updated afterwards - the error says this:
```
## Update the config failed. Please correct.
## Error validating section resotoworker: Kind:resotoworker Property:pool_size is not valid: Expected type int64 but got str: {"collector": ["aws"], "graph": "resoto", "timeout": 10800, "pool_size": "$(POOLSIZE)", "fork_process": true, "graph_merge_kind": "account", "debug_dump_json": true, "tempdir": null, "cleanup": true, "cleanup_pool_size": 16, "cleanup_dry_run": false, "no_tls": false, "web_host": "::", "web_port": 9956, "web_path": "/"}
```

2) The UI does not get the information which path is overridden.

lloesche (changes requested):
The override works great. However the implementation differs from the design doc.
The arg `--override-path` is supposed to be the path to the folder where my override configs are stored in. See design doc for details.

Right now when I edit the config on disk the file is not being re-read on changes. Meaning whenever I change the config on disk I have to restart the component manually instead of the right component restarting or reloading like it does when I edit the config from the CLI.

This has the unfortunate side effect that when I change the config file I also always have to restart the core. And currently if I restart the core but not the worker then the worker still runs with the old config.

I.e. when any of the config files changes on disk it should be reloaded and the core needs to notify the component that their config has changed, so they reload it as well. Same as if I edit the config using `config edit` or the UI.

Setting an environment variable for the worker config on the worker was being picked up by the worker, but setting the same variable on the core did not result in it being substituted.

Right now when config is being overridden, the information is displayed at the top of the config but also in the actual config. The problem here is that I can still edit the value and the edit is being stored but never displayed.

E.g. I have an override
```
resotoworker:
  collector:
    - '$(RESOTO_COLLECTOR)'
```

And then in the config I changed collector to `example`. It was never displayed but it was changed in the database.

The config edit or UI should always show the config that's currently stored in the database  but comment that it's not being active because of an override.

resotocore/resotocore/config/config_handler_service.py
```diff
@@ -125,11 +145,46 @@ async def update_configs_model(self, kinds: List[Kind]) -> Model:
...
+                    }
+                else:
+
+                    def mkstr(val: Any) -> str:
+                        return f'[{", ".join(val)}]' if isinstance(val, list) else str(val)
+
+                    if existing is None:
+                        return f"{mkstr(update)} - new value"
+
+                    return f"{mkstr(update)} - previously was {mkstr(existing)}"
```
aquamatthias, line 169:
```suggestion
                    return f"{mkstr(update)} - overrides existing {mkstr(existing)}"
```

resotocore/resotocore/config/config_handler_service.py
```diff
@@ -125,11 +145,46 @@ async def update_configs_model(self, kinds: List[Kind]) -> Model:
...
+
             yaml_str = ""
+
+            if self.core_config.overrides:
+                yaml_str += (
```
aquamatthias, line 174: That puts the overriding statement on top. Didn't we want to have it on the property level?

resotocore/resotocore/core_config.py
```diff
@@ -622,13 +625,31 @@ def parse_config(args: Namespace, core_config: Json, command_templates: Optional
...
     migrated = migrate_core_config(core_config)
     adjusted = migrated.get(ResotoCoreRoot) or {}
     for path, value in set_from_cmd_line.items():
         if value is not None:
             adjusted = set_value_in_path(value, path, adjusted)
 
+    # merge the file overrides into the adjusted config
+    if all_config_overrides:
+        # here we only care about the resotocore overrides
+        core_overrides = all_config_overrides.get("resotocore", {})
```
aquamatthias, line 647: why is resotocore special here?

resotocore/resotocore/dependencies.py
```diff
@@ -184,6 +185,19 @@ def key_value(kv: str) -> Tuple[str, JsonElement]:
         "Note: this argument allows multiple overrides separated by space. "
         "Example: --override resotocore.api.web_hosts=localhost,some.domain resotocore.api.web_port=12345",
     )
+    parser.add_argument(
+        "--override-path",
+        nargs="+",
+        type=Path,
```
aquamatthias, line 191: It allows only files, not directories - idea was to have a directory with all my config overrides.

aquamatthias (commented):
Thanks. Did you push all local changes? A couple of points were already reported the last time - maybe some old version?

I tried to set this value in my config (resoto core):
```
    max_request_size: $(REQUEST_SIZE)
```
which gives an error: `Property:max_request_size is not valid: Expected type int64 but got str:` - shouldn't this work?

resotocore/resotocore/config/config_handler_service.py
meln1k: That's a bit more complicated and currently being worked on by me.

resotocore/resotocore/core_config.py
meln1k: that's the core only part of the config.

aquamatthias (commented):
Thanks. Please find my comments.

I ran into 3 bugs from my POV:

1) I defined a config element as env var:
```yaml
resotocore:
    max_request_size: $(MAXSIZE)
```
and got this error:
```
## Update the config failed. Please correct.
## Error executing task: Validation Errors:
## api:
## - max_request_size:
##   - must be of integer type
```

2) I changed the configuration of the worker. This works fine. But the override became the default configuration: after the override was removed, the config still uses the overide values.

3) After I defined an override for resoto.worker, the order of the config file entries was different. Would be great if the order of config entries is stable.

resotocore/resotocore/config/config_override_service.py
```diff
@@ -0,0 +1,133 @@
...
+        self.stop_watcher = asyncio.Event()
+        self.override_change_hooks: List[Callable[[Dict[ConfigId, Json]], Awaitable[Any]]] = []
+        self.watcher_task: Optional[asyncio.Task[Any]] = None
+        self.overrides = self._load_overrides()
+        self.mtime_hash: int = 0
+
+    def add_override_change_hook(self, hook: Callable[[Dict[ConfigId, Json]], Awaitable[Any]]) -> None:
+        self.override_change_hooks.append(hook)
+
+    # no async here because pagacache will keep this in memory anyway
```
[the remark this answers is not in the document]
meln1k: a typo in page cache

resotocore/requirements-test.txt
```diff
@@ -3,6 +3,6 @@ pytest==7.2.1
 pytest-cov==4.0.0
 pytest-runner==6.0.0
 pytest-asyncio==0.20.3
-deepdiff==6.2.3
 hypothesis==6.68.1
 resotoclient==1.2.1
+watchfiles==0.18.1
```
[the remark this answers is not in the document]
meln1k: Nope, thanks for noticing

resotocore/resotocore/config/config_override_service.py
```diff
@@ -0,0 +1,133 @@
...
+                        await hook(self.overrides)
+
+                await asyncio.sleep(self.sleep_time)
+
+        # watcher is already running
+        if self.watcher_task:
+            return
+
+        self.stop_watcher.clear()
+        self.watcher_task = asyncio.create_task(watcher())
```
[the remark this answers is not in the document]
meln1k: Periodic is good for sync tasks, but here we use a lot of asyncs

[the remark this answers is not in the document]
aquamatthias: It is designed for async tasks - see usages.

resotocore/resotocore/config/config_handler_service.py
```diff
@@ -66,11 +78,49 @@ async def coerce_and_check_model(self, cfg_id: ConfigId, config: Json, validate:
         # If we come here, everything is fine
         return final_config
 
+    async def coerce_config(self, config: Json) -> Json:
```
aquamatthias, line 81: why does `coerce_and_check_model` does not use `coerce_config` - please DRY it up.

resotocore/resotocore/__main__.py
```diff
@@ -194,7 +205,9 @@ async def on_start() -> None:
         await cli.start()
         await inspector.start()
         await task_handler.start()
+        await config_handler.start()
         await core_config_handler.start()
+        config_override_service.watch_for_changes()
```
aquamatthias, line 210: consider to make it async and call it start / stop respectively

resotocore/resotocore/config/config_override_service.py
```diff
@@ -0,0 +1,133 @@
...
+                            ]
+                        )
+                    else:
+                        config_files.append(path)
+
+                # a quick optimization to avoid reading the files if none has been changed
+                mtime_hash = 0
+                config_files = sorted(config_files)
+                for file in config_files:
+                    mtime_hash = hash((mtime_hash, (await aos.stat(file)).st_mtime))
```
aquamatthias, line 106: @lloesche I remember relying on mtime is not always safe - since ram file systems do not maintain that information? Not sure how k8s is handling it with configmaps. wdyt?

resotocore/resotocore/config/config_override_service.py
```diff
@@ -0,0 +1,133 @@
...
+
+                if mtime_hash == self.mtime_hash:
+                    await asyncio.sleep(self.sleep_time)
+                    continue
+                self.mtime_hash = mtime_hash
+
+                # fine to block here since it happens once in a lifetime
```
aquamatthias, line 113: I think the comment is wrong inside the while loop?

resotocore/resotocore/config/config_override_service.py
```diff
@@ -0,0 +1,133 @@
...
+                    [file for file in path.iterdir() if file.is_file() and file.suffix in (".yml", ".yaml")]
+                )
+            else:
+                config_files.append(path)
+
+        # json with all merged overrides for all components such as resotocore, resotoworker, etc.
+        overrides_json: Json = {}
+        # merge all provided overrides into a single object, preferring the values from the last override
+        for config_file in config_files:
+            with config_file.open() as f:
```
aquamatthias, line 50: Consider using aiofiles for reading the files.

resotocore/resotocore/config/config_handler_service.py
```diff
@@ -125,15 +160,37 @@ async def update_configs_model(self, kinds: List[Kind]) -> Model:
         return updated
 
     async def config_yaml(self, cfg_id: ConfigId, revision: bool = False) -> Optional[str]:
-        config = await self.get_config(cfg_id)
+        config = await self.get_config(cfg_id, apply_overrides=False, resolve_env_vars=False)
```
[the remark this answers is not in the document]
meln1k: done

resotocore/resotocore/config/config_override_service.py
meln1k: usually this part is never reachable due to the mtime shortcut above

resotocore/resotocore/config/config_override_service.py
meln1k: I wanted to keep it like that for the reason of getting a fully constructed object after calling init.

resotocore/resotocore/config/config_handler_service.py
meln1k: I'm not sure that would help because in coerce_and_check_model coercion is done implicitly via the `check_valid` method, and DRYing it would make the whole thing less readable.

aquamatthias: Missed that point - makes sense.

resotocore/resotocore/config/config_handler_service.py
```diff
@@ -66,11 +78,51 @@ async def coerce_and_check_model(self, cfg_id: ConfigId, config: Json, validate:
...
+        final_config = {}
+        for key, value in config.items():
+            if key in model:
+                value_kind = model[key]
+                coerced = value_kind.coerce(value)
+                sorted_conf = value_kind.sort_json(coerced) if isinstance(coerced, dict) else coerced
+                final_config[key] = sorted_conf
+            else:
+                final_config[key] = value
+        # If we come here, everything is fine
```
aquamatthias, line 93: This comment does not make sense here.
```suggestion
```

resotocore/resotocore/config/config_handler_service.py
```diff
@@ -66,11 +78,51 @@ async def coerce_and_check_model(self, cfg_id: ConfigId, config: Json, validate:
...
+        if resolve_env_vars:
+            resolved_conf = {k: replace_env_vars(v, os.environ) for k, v in updated_conf.items()}
+            coerced_conf = await self.coerce_config(resolved_conf)
+            updated_conf = coerced_conf
+
+        new_ce = attrs.evolve(
+            conf,
+            config=updated_conf,
+        )
+        return new_ce
```
aquamatthias, line 125:
```suggestion
        return attrs.evolve(conf, config=updated_conf)
```

resotocore/resotocore/config/config_override_service.py
```diff
@@ -0,0 +1,133 @@
...
+                for file in config_files:
+                    mtime_hash = hash((mtime_hash, (await aos.stat(file)).st_mtime))
+
+                if mtime_hash == self.mtime_hash:
+                    await asyncio.sleep(self.sleep_time)
+                    continue
+                self.mtime_hash = mtime_hash
+
+                # fine to block here since it happens once in a lifetime
+                overrides = self._load_overrides(silent=True)
```
aquamatthias, line 114: consider making it run async
```suggestion
                overrides = await run_async(self._load_overrides, silent=True)
```

resotocore/resotocore/config/config_override_service.py
aquamatthias: See comment on the next line - you can make it async. 
Still would remove this comment, since it is misleading.
```suggestion
```

resotocore/resotocore/core_config.py
```diff
@@ -81,10 +82,25 @@ def default_hosts() -> List[str]:
     return ["0.0.0.0"] if inside_docker() else ["localhost"]
 
 
+def strip_env_vars_paths(config: JsonElement) -> JsonElement:
```
aquamatthias, line 85: consider to define it inside `validate_config` - not used anywhere else.

resotocore/resotocore/core_config.py
```diff
@@ -646,16 +667,26 @@ def parse_config(args: Namespace, core_config: Json, command_templates: Optional
...
-        log.warning(f"Can not adjust configuration: {e}", exc_info=e)
+    # here we only care about the resotocore overrides
+    core_config_overrides = (get_core_overrides() or {}).get(ResotoCoreRoot)
+    # merge the file overrides into the adjusted config
+    if core_config_overrides:
+        adjusted = merge_json_elements(adjusted, core_config_overrides)
 
     try:
+        # replacing the env vars and exploding if there is no value
+        adjusted = replace_env_vars(adjusted, os.environ, ignore_missing=False)
```
aquamatthias, line 678: The exploding part will be catched by the surrounding try/catch block.
The result will be the default configuration. I guess this error should bubble up?
Example:
```
23-02-28 08:33:21|ERROR|Final configuration can not be parsed! Fall back to default configuration.
ValueError: Env var not found: NOT_PROVIDED_ENV_VAR
```

aquamatthias (commented):
Thanks. The following things have been tested:
- define overrides and see them reflected in the config
- changing the config keeps the order of keys stable
- defining env vars for kinds other than string seems to work
- changing the file inside the override directory will yield a config change event

I tried the following and ran into an error. I defined the override for resotocore like this:
```
resotocore: 
  api: 
    max_request_size: test
```

The config is loaded without errors and the client code that uses the config breaks.
A validation error should have been thrown in this case, or?

resotocore/resotocore/__main__.py
```diff
@@ -107,18 +114,41 @@ def run_process(args: Namespace) -> None:
             warnings.simplefilter("ignore", HTTPWarning)
             # wait here for an initial connection to the database before we continue. blocking!
             created, system_data, sdb = DbAccess.connect(args, timedelta(seconds=120), verify=False)
-            config = config_from_db(args, sdb)
+
+            # create a tiny version of config override service that can validate the core config
+            async def init_core_config_override() -> ConfigOverrideService:
```
aquamatthias, line 119: Consider to move this code to a minimal `ConfigOverride` outside of main.

lloesche (approved):
This is great! 🚢

PR #1445, opened 2023-02-14
[ci][chore] Write AWS policy JSON to resoto.com repo

TheCatLady:
# Description

The AWS policy JSON cannot be fetched at build time from Netlify's build environment due to some networking/proxy issues on their side, so just write the policy JSON to the `someengineering/resoto.com` repo directly.

This also has the benefit of builds being able to work locally without an internet connection, though the downside is that we'll be maintaining two copies of the JSON (one on our DigitalOcean CDN and another on GitHub).

See also: someengineering/resoto.com#609

# Code of Conduct

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -12,20 +12,17 @@ on:
       - "docker/**"
       - "dockerV2/**"
       - ".github/workflows/docker-build.yml"
-  workflow_dispatch:
 
 jobs:
   split-build:
-    name: Build split Docker images
+    name: "Build split Docker images" # Do not rename without updating workflow defined in publish.yml
     runs-on: ubuntu-latest
 
     steps:
       - name: Check out repository
-        if: github.event_name != 'workflow_dispatch'
         uses: actions/checkout@v3
 
       - name: Get short commit SHA
-        if: github.event_name != 'workflow_dispatch'
         id: sha
         run: echo "short=${GITHUB_SHA::7}" >> $GITHUB_OUTPUT
 
@@ -46,7 +43,6 @@ jobs:
           free -m
 
       - name: Set build platforms
-        if: github.event_name != 'workflow_dispatch'
         id: platform
         run: |
           GITHUB_REF="${{ github.ref }}"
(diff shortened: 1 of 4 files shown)


aquamatthias (approved):
thanks

.github/workflows/docker-build.yml
```diff
@@ -425,22 +433,31 @@ jobs:
...
-          cp resotocore/resotocore/static/api-doc.yaml resoto.com/openapi/resotocore-${{steps.release.outputs.docsVersion}}.yml
-          echo $(jq '.["${{steps.release.outputs.docsVersion}}"].version="${{ steps.release.outputs.tag }}" | .["${{steps.release.outputs.docsVersion}}"].link="${{ steps.release_notes.outputs.link }}"' resoto.com/latestRelease.json) > resoto.com/latestRelease.json
+          cp resotocore/resotocore/static/api-doc.yaml resoto.com/openapi/resotocore-${{ steps.release.outputs.docsVersion }}.yml
+          echo $(jq '.["${{ steps.release.outputs.docsVersion }}"].version="${{ steps.release.outputs.tag }}" | .["${{ steps.release.outputs.docsVersion }}"].link="${{ steps.release_notes.outputs.link }}"' resoto.com/latestRelease.json) > resoto.com/latestRelease.json
+
+      - name: Update AWS policy JSON
+        if: steps.release.outputs.prerelease == 'false'
+        shell: bash
+        working-directory: ./resoto.com/aws/${{ steps.release.outputs.docsVersion }}
+        run: |
```
lloesche, line 451: There is a race condition here between the AWS plugin workflow uploading and this workflow downloading the JSON files. Right now there's a good chance it will work, because the Docker build takes longer than the PIP build. But there's not guarantee for this code working.

TheCatLady: Yeah, I was thinking about how to best handle this.

We had discussed splitting the release/docs logic into a separate workflow previously, so I think this is a good time to make that change. See my latest commit :)

TheCatLady:
@aquamatthias @lloesche These changes have been tested in the `someengineering/playground` repo, please take a look!

aquamatthias (approved):
Thanks!

PR #1466, opened 2023-02-28
[gcp][feat] handling of custom machine types for Gcp Instances

anjafr:
# Description

- makes sure that for each unique custom `GcpMachineType` of `GcpInstance`s one node gets added to the graph. All `GcpInstance`s copy the information about cores, memory and machine type from the actual machine type onto themselves
- introduces a `FixturedClient` for Gcp Testing that allows to set custom attributes
- introduces async capabilities for collection (same `submit_work` logic as in the AWS collector) [this reduced collection time of compute resources for an example project on my local machine from almost 2 minutes down to 30 seconds]

# To-Dos

# Code of Conduct


plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2713,10 +2719,43 @@ class GcpInstance(GcpResource, BaseInstance):
...
+                id=source["machineType"].split("/")[-1],
+                tags={},
+                # TODO figure out how this works?!
+                # _zone=self.zone(graph),
+                # _account=self.account(graph),
+                link=source["machineType"],
+            )
+            machine_type.instance_cores = float(id.split("-")[-2])
+            mem = int(id.split("-")[-1])
+            machine_type.instance_memory = float(mem / 1024)
```
aquamatthias, line 2751: why not make it part of the ctor?

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2713,10 +2719,43 @@ class GcpInstance(GcpResource, BaseInstance):
     instance_start_restricted: Optional[bool] = field(default=None)
     instance_status_message: Optional[str] = field(default=None)
     instance_tags: Optional[GcpTags] = field(default=None)
+    instance_type: Optional[str] = None
+    instance_cores: float = 0.0
+    instance_memory: float = 0.0
+    _machine_type_link: Optional[str] = field(default=None, alias="_machine_type_link")
+    _machine_type: Optional[BaseInstanceType] = field(default=None, alias="_machine_type")
```
aquamatthias, line 2726: why do we need this - can we remove it?

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2713,10 +2719,43 @@ class GcpInstance(GcpResource, BaseInstance):
     instance_start_restricted: Optional[bool] = field(default=None)
     instance_status_message: Optional[str] = field(default=None)
     instance_tags: Optional[GcpTags] = field(default=None)
+    instance_type: Optional[str] = None
+    instance_cores: float = 0.0
+    instance_memory: float = 0.0
```
aquamatthias, line 2724: Already defined in BaseInstance?

plugins/gcp/resoto_plugin_gcp/resources/compute.py
anjafr: It does not make sense to check if `self._machine_type` is an instance of `BaseInstanceType` when this property is not there/not this.
See [here](https://github.com/someengineering/resoto/blob/5987d3678789d4a6c48be60ec5345baac873dbd3/plugins/gcp/resoto_plugin_gcp/resources/compute.py#L2728-L2733): This is the part we agreed on keeping because it defines the `BaseInstance`

anjafr:
@aquamatthias still needs implementing of a `client.get()` kinda method, equivalent to the aws one. Do you think this makes more overall sense now? No more keeping `_machine_type` around!

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2714,13 +2720,49 @@ class GcpInstance(GcpResource, BaseInstance):
...
+                gcp_service="compute",
+                action="get",
+                resource="machineType",
+                project=graph_builder.project,
+                zone=zone,
+                machineType=name,
+            )
+            return GcpMachineType.from_api(result)
+
+        machine_type = find_existing_machine_type() or create_adhoc_machine_type()
```
aquamatthias, line 2746: The current collector detects a custom machine type by this line `if resource.instance_type in ("", None) and "custom" in resource._machine_type_link`.
Plus: you need to make sure to not create multiple instances for the same custom machine type. 
After collecting all machine instances, you know all custom machine types.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2714,13 +2720,49 @@ class GcpInstance(GcpResource, BaseInstance):
     instance_status_message: Optional[str] = field(default=None)
     instance_tags: Optional[GcpTags] = field(default=None)
 
+    def set_machine_type(self, machine_type_link_ish: str, graph_builder: GraphBuilder) -> None:
+        def find_existing_machine_type() -> Optional[GcpMachineType]:
```
aquamatthias, line 2724: If you call this method in `post_process` than it is not safe that the related machine type is already available.
Maybe it is better to copy the relevant data in the `connect_in_graph` phase.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2714,13 +2720,49 @@ class GcpInstance(GcpResource, BaseInstance):
...
+            if not machine_type_link_ish.startswith("https://"):
+                machine_type_link_ish = (
+                    f"https://www.googleapis.com/compute/v1/projects/{graph_builder.project.id}/{machine_type_link_ish}"
+                )
+            return graph_builder.node(clazz=GcpMachineType, link=machine_type_link_ish)
+
+        # TODO: Maybe as class method on GcpMachineType?
```
aquamatthias, line 2731: sounds good

plugins/gcp/resoto_plugin_gcp/resources/compute.py
anjafr: Okay, agreed. We would need to put all this logic in a step that happens after all collection has been completed in order to work as intended. Putting it in `connect_in_graph()` would do that, however that would be wrong semantically, as it leaves `GcpInstance` in an incorrect / unfinished state after `post_processing()`.
If we're OK with that, I'll move it downwards.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
anjafr: >  Plus: you need to make sure to not create multiple instances for the same custom machine type.

Sorry, what do you mean? We are relying on the "if not found then create adhoc", the create-part of which would only happen for the very first `GcpInstance` of each (project, zonal) `GcpMachineType`.

aquamatthias: Making requests to GCP happens asynchronously, while the list of resources is walked synchronously.
Waling 2 instances with the same custom instance type could yield fetching the same instance type twice.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
aquamatthias: The only action we would do in this phase: searching the related machine type and copy cores/memory information down to every instance.
Not ideal I agree, but I think it is better this way as to introduce dependencies to make sure instance types are available before instances are fetched.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -16,6 +18,8 @@
 from resotolib.json_bender import Bender, S, Bend, ForallBend, MapDict, MapValue, F
 from resotolib.types import Json
 
+log = resotolib.logger.getLogger("resoto." + __name__)
```
aquamatthias, line 21:
```suggestion
log = logging.getLogger("resoto.plugins.gcp")
```

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -1006,6 +1010,7 @@ class GcpFirewallPolicyRule:
     target_service_accounts: Optional[List[str]] = field(default=None)
 
 
+# TODO Firewall Policies are on org level, parentId is org id or folder id
```
aquamatthias, line 1013: They exist on project level and *also* on org level.
We should fetch the one on project level as part of the project collector.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2715,18 +2721,48 @@ class GcpInstance(GcpResource, BaseInstance):
...
+        # - extract machineType if custom
+        # - then create GcpMachineType object for each unique custom machine type
+        # - add the new objects to the graph
+        result = super().collect(raw, builder)
+        custom_machine_types = list(
+            set([instance.instance_machine_type for instance in result if "custom" in instance.instance_machine_type])
+        )
+        for machine_type in custom_machine_types:
+            zone = machine_type.split("/")[-3]
+            name = machine_type.split("/")[-1]
```
aquamatthias, line 2761: a) you do not need to split twice - assign the variable and get the name and zone
b) can you add a comment with an example: hard to see if the access is correct

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -2715,18 +2721,48 @@ class GcpInstance(GcpResource, BaseInstance):
...
+            GcpMachineType.collect_individual(builder, zone, name)
```
aquamatthias, line 2762: Can we load this asynchronously?

plugins/gcp/test/random_client.py
```diff
@@ -64,14 +64,19 @@ def random_datetime() -> str:
 # dictionary keys under .items (used for aggregated list zone result) -> return zone ids
 PredefinedDictKeys = {".items": [a.id for a in random_zones]}
 
+# type_name -> property_name -> parameter_name
```
aquamatthias, line 67: I guess this comment is wrong...

plugins/gcp/test/test_compute.py
```diff
@@ -103,7 +103,32 @@ def test_gcp_instance_template(random_builder: GraphBuilder) -> None:
...
+    CUSTOM_MACHINE_TYPE_FULL = (
+        f"https://www.googleapis.com/compute/v1/projects/{random_builder.project.id}/{CUSTOM_MACHINE_TYPE_PART}"
+    )
+    fixture_replies = {
+        "Instance": {"machineType": lambda: CUSTOM_MACHINE_TYPE_PART},
+        "MachineType": {"selfLink": lambda: CUSTOM_MACHINE_TYPE_FULL},
+    }
+    assert len(random_builder.resources_of(GcpMachineType)) == 0
+
+    with FixturedClient(random_builder, fixture_replies) as random_builder:
```
aquamatthias, line 122: Nice!

aquamatthias (approved):
Cool! 2 nits - otherwise looks great to me.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
anjafr: yes we should - in another PR 👍🏻

aquamatthias (approved):
Going forward, we want to consolidate shared classes like ExecutorQueue into the library.
Since there is also consolidation on the AWS side, let's defer this. 
Nice

PR #1505, opened 2023-03-16
[resotocore][feat] Allow metadata for kinds.

aquamatthias:
# Description

- Allow to define metadata as part of the resource class
- Allow getting a flattened model
- Add icon and group to all base resources

# To-Dos

# Code of Conduct

### DIFF
--- a/resotolib/resotolib/baseresources.py
+++ b/resotolib/resotolib/baseresources.py
@@ -7,7 +7,7 @@ import hashlib
 import weakref
 from resotolib.logger import log
 from enum import Enum
-from typing import Dict, Iterator, List, ClassVar, Optional, TypedDict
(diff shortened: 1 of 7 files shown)


resotolib/resotolib/baseresources.py
```diff
@@ -758,6 +771,7 @@ class BaseNetworkShare(BaseResource, ABC):
 @define(eq=False, slots=False)
 class BaseVolume(BaseResource):
```
lloesche, line 772: Given that types don't mean anything in terms of cost and actual resources do cost money, making volume type and volume the same icon is confusing. I would make `*_type` something like a text/description icon as the type is just metadata and compute and storage some distinct icons as you actually pay for them.
I wonder if types need a distinct icon at all... they are mostly uninteresting and could have some generic default icon. I don't even think I would want them shown in most diagrams.

PR #1518, opened 2023-03-23
[resoto][feat] auto enablement of collector plugins

lloesche:
# Description

This PR implements auto enablement of collector plugins.

A `BaseCollectorPlugin` can now implement an `auto_enabled()` method that returns a bool whether the collector should be enabled or not. The default for all collectors is `False`.

This PR also implements `auto_enabled()` for the AWS collector.

# Code of Conduct

### DIFF
--- a/resotolib/resotolib/baseplugin.py
+++ b/resotolib/resotolib/baseplugin.py
@@ -215,6 +215,11 @@ class BaseCollectorPlugin(BasePlugin):
         pass
 
     @staticmethod
+    def auto_enabled() -> bool:
+        """Should this collector be enabled by default?"""
+        return False
+
+    @staticmethod
     def update_tag(config: Config, resource: BaseResource, key: str, value: str) -> bool:
         """Update the tag of a resource"""
         return resource.update_tag(key, value)
--- a/resotolib/resotolib/baseplugin.py
+++ b/resotolib/resotolib/baseplugin.py
@@ -215,7 +215,7 @@ class BaseCollectorPlugin(BasePlugin):
         pass
 
     @staticmethod
-    def auto_enabled() -> bool:
+    def auto_enableable() -> bool:
         """Should this collector be enabled by default?"""
         return False
 
--- a/resotoworker/resotoworker/config.py
+++ b/resotoworker/resotoworker/config.py
@@ -1,4 +1,4 @@
-import concurrent.futures
+from concurrent.futures import ThreadPoolExecutor, TimeoutError
 from collections import namedtuple
 from resotolib.graph import GraphMergeKind
 from resotolib.config import Config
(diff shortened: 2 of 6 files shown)


resotoworker/resotoworker/config.py
```diff
@@ -1,19 +1,47 @@
...
+from resotolib.baseplugin import BaseCollectorPlugin
+from resotolib.logger import log
 from attrs import define, field
-from typing import ClassVar, Optional, List
+from typing import ClassVar, Optional, List, Type
 
 
-def add_config(config: Config) -> None:
+_collector_plugins: List[Type[BaseCollectorPlugin]] = []
+PluginAutoEnabledResult = namedtuple("PluginAutoEnabledResult", ["cloud", "auto_enabled"])
```
meln1k, line 13: What do you think about a frozen attrs class instead? e.g.

```python
@frozen
class PluginAutoEnabledResult:
    cloud: str
    auto_enabled: bool
```

resotoworker/resotoworker/config.py
```diff
@@ -1,19 +1,47 @@
...
+
+
+def add_config(config: Config, collector_plugins: List[Type[BaseCollectorPlugin]]) -> None:
+    global _collector_plugins
+    _collector_plugins = collector_plugins
```
meln1k, line 18: Maybe it is worth moving this to a separate method and calling it before `add_config`?

resotoworker/resotoworker/config.py
```diff
@@ -1,19 +1,47 @@
...
+        log.error(f"Unhandled exception while getting auto-enabled collectors: {e}")
+    return []
+
+
 @define
 class ResotoWorkerConfig:
     kind: ClassVar[str] = "resotoworker"
     collector: List[str] = field(
-        factory=lambda: ["example"],
+        factory=get_default_collectors,
```
meln1k, line 44: An idea: we can cache the result of get_default_collectors call in a separate value, and use it here. E.g.
```python
default_collectors = get_default_collectors()
...
@define
class ResotoWorkerConfig:
    kind: ClassVar[str] = "resotoworker"
    collector: List[str] = field(
        factory=lambda: default_collectors,
...
```

This way we don't have to spawn a thread pool every time we want to init the worker config.

resotolib/resotolib/baseplugin.py
```diff
@@ -214,6 +214,11 @@ def collect(self) -> None:
         """Collects all the Cloud Resources"""
         pass
 
+    @staticmethod
+    def auto_enabled() -> bool:
```
meln1k, line 218: It feels like we could use a different name. What about `ready_to_start`?

meln1k (commented):
a couple of comments

resotoworker/resotoworker/config.py
lloesche: I think add_config is the right place, because we want to ensure that it happens before the config is added. If I move it somewhere else there's a chance for misuse. If it's in here there's no chance it doesn't happen before the config is added.

resotoworker/resotoworker/config.py
lloesche: Great idea, will do!

resotolib/resotolib/baseplugin.py
lloesche: Idk, the method determines whether the collector will be automatically enabled. It doesn't say anything about whether the collector is ready it also doesn't configure the collector. All it does is automatically add the collector name to the list of enabled collectors. Hence the name auto_enabled(). Of the suggestions auto_enableable would fit, but enableable is also not a common word you'd find in most dictionaries. What about `auto_ activatable`?

meln1k (approved):

PR #1519, opened 2023-03-23
[gcp][fix] Ensure proper handling of region and zone for all resources

anjafr:
# Description

# To-Dos

# Code of Conduct


plugins/gcp/resoto_plugin_gcp/resources/base.py
```diff
@@ -168,29 +168,28 @@ def add_node(self, node: GcpResourceType, source: Optional[Json] = None) -> Opti
         log.debug(f"{self.name}: add node {node}")
         node._cloud = self.cloud
         node._account = self.project
-        if source is not None and InternalZoneProp in source:
-            if zone := self.zone_by_name.get(source[InternalZoneProp]):
-                node._zone = zone
-                node._region = self.region_by_zone_name[source[InternalZoneProp]]
-                self.add_edge(node, node=zone, reverse=True)
+        if not (node._region or isinstance(node, GcpRegion)):
```
aquamatthias, line 171: - Why wouldn't we link a GcpRegion to a project? (or where does this happen)?
- If the resource already has a region, where is the edge created?

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -3292,6 +3292,8 @@ def collect_individual(cls: Type[GcpResource], builder: GraphBuilder, zone: str,
             machineType=name,
         )
         machine_type_obj = GcpMachineType.from_api(result)
+        machine_type_obj._zone = builder.zone_by_name[zone]
+        machine_type_obj._region = builder.region_by_zone_name[zone]
```
aquamatthias, line 3296: So we collect machine types for a specific zone. To reflect the client behaviour for zonal responses, can't we simply do:
```python
result[InternalZoneProp]=zone
```
This should take care for setting the relevant properties and adding edges.

aquamatthias (approved):
Nice! One question if we should inline the function.

plugins/gcp/resoto_plugin_gcp/resources/base.py
```diff
@@ -168,33 +168,53 @@ def add_node(self, node: GcpResourceType, source: Optional[Json] = None) -> Opti
...
-            # log.error(f"Neither zone nor region is set for node {source}, add to project.")
-            self.add_edge(node, node=self.project)
+
+        self.standard_edges(node, source)
+
         with self.graph_nodes_access:
             self.graph.add_node(node, source=source or {})
         return node
 
+    def standard_edges(self, node: GcpResourceType, source: Optional[Json] = None) -> None:
```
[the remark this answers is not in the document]
anjafr: Its only use case is within `add_node`. I've moved the functionality into its own method mostly for testability reasons. There is a new test case that creates a blank `GcpProjectCollector`, injects a graph using some fixtures and then calls the new `standard_edges` method, without going through the full `collect` call.

Not sure how we'd keep that way of testing with the inlined function, but happy for any pointers / ideas!

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -3291,6 +3291,7 @@ def collect_individual(cls: Type[GcpResource], builder: GraphBuilder, zone: str,
             zone=zone,
             machineType=name,
         )
+        result[InternalZoneProp] = zone
```
aquamatthias, line 3294: Please add a comment explaining this line of code. It is not obvious.

aquamatthias (approved):
thanks

plugins/gcp/resoto_plugin_gcp/resources/compute.py
anjafr: will do

PR #1520, opened 2023-03-24
[resotoworker][feat] Save files to home dir via config

meln1k:
# Description

Added a feature that allows saving files to the resotoworker's home directory by defining them in the configuration.
# To-Dos

# Issues Fixed

- Fixes #XXXX

# Code of Conduct

### DIFF
--- a/resotoworker/resotoworker/__main__.py
+++ b/resotoworker/resotoworker/__main__.py
@@ -34,6 +34,7 @@ from resotolib.web.metrics import WebApp
 from resotoworker.cleanup import cleanup
 from resotoworker.collect import Collector
 from resotoworker.config import add_config
+from resotoworker.utils import write_files_to_home_dir, write_utf8_file
 from resotoworker.pluginloader import PluginLoader
 from resotoworker.resotocore import Resotocore
 from resotoworker.tag import core_tag_tasks_processor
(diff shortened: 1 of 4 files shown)


aquamatthias (dismissed):
Nice!

resotoworker/resotoworker/config.py
```diff
@@ -9,6 +9,13 @@ def add_config(config: Config) -> None:
     config.add_config(ResotoWorkerConfig)
 
 
+@define
+class HomeDirectoryFile:
+    kind: ClassVar[str] = "resotoworker_home_directory_file"
```
lloesche, line 14: This should be part of resotolib and can be used by anything, not just resotoworker.

resotoworker/resotoworker/__main__.py
```diff
@@ -97,6 +98,8 @@ def main() -> None:
     plugin_loader.add_plugin_config(config)
     config.load_config()
 
+    write_files_to_home_dir(config.resotoworker.write_files_to_home_dir, write_utf8_file)
```
lloesche, line 101: Can't this be part of each individual file object? Instead of having to call the write in the worker explicitly. Whenever I instantiate one of those HomedirectoryFile objects, can't they check if they need to write the file contents?

resotoworker/resotoworker/config.py
meln1k: That's a great idea for the future, when we have a need to use this feature not only in the resotoworker but also in other components. But let's keep it simple for now. We can always move it to resotolib when the need is there.

aquamatthias (approved):

PR #1523, opened 2023-03-29
[gcp][feat] OnDemand cost information for GcpDiskType

anjafr:
# Description

OnDemand cost information for GcpDiskType

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/gcp/resoto_plugin_gcp/resources/compute.py
+++ b/plugins/gcp/resoto_plugin_gcp/resources/compute.py
@@ -767,6 +767,57 @@ class GcpDiskType(GcpResource, BaseVolumeType):
     default_disk_size_gb: Optional[str] = field(default=None)
     valid_disk_size: Optional[str] = field(default=None)
 
+    def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
+        """Adds edges from disk_types type to SKUs and determines ondemand pricing"""
+        if not self.name:
+            return
+
+        log.debug((f"Looking up pricing for {self.rtdname} in {self.region().rtdname}"))
+        resource_group_map = {
+            "local-ssd": "LocalSSD",
+            "pd-balanced": "SSD",
+            "pd-ssd": "SSD",
+            "pd-standard": "PDStandard",
+        }
+        resource_group = resource_group_map.get(self.name)
+
+        def filter(sku: GcpSku) -> bool:
+            if not self.name:
+                return False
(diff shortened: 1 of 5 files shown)


plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -767,6 +767,57 @@ class GcpDiskType(GcpResource, BaseVolumeType):
...
+        log.debug((f"Looking up pricing for {self.rtdname} in {self.region().rtdname}"))
+        resource_group_map = {
+            "local-ssd": "LocalSSD",
+            "pd-balanced": "SSD",
+            "pd-ssd": "SSD",
+            "pd-standard": "PDStandard",
+        }
+        resource_group = resource_group_map.get(self.name)
+
+        def filter(sku: GcpSku) -> bool:
```
aquamatthias, line 784: filter is a python predefined name. please choose another one.

plugins/gcp/resoto_plugin_gcp/resources/compute.py
```diff
@@ -767,6 +767,58 @@ class GcpDiskType(GcpResource, BaseVolumeType):
     default_disk_size_gb: Optional[str] = field(default=None)
     valid_disk_size: Optional[str] = field(default=None)
 
+    RESOURCE_GROUP_MAP = {
```
aquamatthias, line 770: This is defined in a data class - it becomes an attribute of every instance if not defined as classvar explicitly.

lloesche:
The current collector has a logic
```
if resource.region(graph).name == "undefined" and resource.zone(graph).name == "undefined":
    log.debug(f"Resource {resource.rtdname} has no region or zone" " - removing from graph")
    graph.remove_node(resource)
    return
```

Is this still implemented somewhere?

anjafr:
> Is this still implemented somewhere?

@lloesche Not in this form, no. We do remove DiskType nodes when they don't have any successors, i.e. aren't used by any actual disk. The new collector defaults to create edges to project if region or zone can't be determined for any resource. I'm not sure on what the initial intention was with the above code. Ultimately we do not keep any unused or otherwise dangling DiskType nodes around.

aquamatthias (approved):
Thanks. Regarding the question from Lukas: all resources that do not belong to a zone or region are put under the project level. If there are resources that we should not collect, we should do it more explicitly than missing properties.

PR #1528, opened 2023-03-30
[gcp][feat] Use next gen collector

anjafr:
# Description
WIP

# To-Dos

## Breaking
- [ ] service names changed for `collect` and `no_collect` config settings
- [ ] ...

# Issues Fixed

- Fixes #XXXX

# Code of Conduct


anjafr:
collection result of current collector:
```
Sent 7908/7908, 3868 nodes, 4040 edges and 0 deferred edges in 0.1121s
Collect ran for 207 seconds
> search is(gcp_resource) | count kind
gcp_project: 1
gcp_bucket: 1
gcp_disk_type: 1
gcp_network: 1
gcp_service: 1
gcp_disk: 2
gcp_machine_type: 2
gcp_instance: 2
gcp_service_sku: 3
gcp_region: 37
gcp_subnetwork: 38
gcp_route: 39
gcp_zone: 112
gcp_quota: 3626
total matched: 3866
total unmatched: 0
```

collection result of new collector:
```
Sent 11335/11335, 5579 nodes, 5756 edges and 0 deferred edges in 0.1890s
Collect ran for 162 seconds
> search is(gcp_resource) | count kind
gcp_project: 1
gcp_disk_type: 1
gcp_network: 1
gcp_disk: 2
gcp_instance: 2
gcp_machine_type: 2
gcp_sku: 4
gcp_firewall: 4
gcp_region: 37
gcp_subnetwork: 38
gcp_route: 39
gcp_zone: 112
gcp_accelerator_type: 212
gcp_interconnect_location: 294
gcp_sql_flag: 562
gcp_node_type: 598
gcp_quota: 3626
total matched: 5535
total unmatched: 0
```

To Do:
- gcp_bucket isn't implemented yet
- tagging and deletion methods
- ...

plugins/gcp/resoto_plugin_gcp/__init__.py
```diff
@@ -118,7 +123,7 @@ def collect_project(
 
         try:
             core_feedback.progress_done(project_id, 0, 1)
-            gpc = GCPProjectCollector(project)
+            gpc = GcpProjectCollector(Config.gcp, Cloud(id="Gcp", name="Gcp"), project, core_feedback)
```
lloesche, line 126: Every collector has `self.cloud` which contains an instance of `Cloud()`.

anjafr: @lloesche Did you mean "every colelctor plugin" 🤔 ? Because this is where the project collector is getting their "self".cloud.

anjafr:
quick update: migrated the `update_tag` logic to the new collectors architecture using the `GcpClient`. So far, works on disks but haven't tested other types of resources yet. 

image

plugins/gcp/resoto_plugin_gcp/resources/base.py
```diff
@@ -301,14 +302,22 @@ class GcpResource(BaseResource):
     link: Optional[str] = None
     label_fingerprint: Optional[str] = None
 
+    @property
+    def _get_identifier(self) -> str:
+        return self.api_spec.accessors[-1][:-1] # Poor persons `singularize(), i.e. ["vpnTunnels"] -> "vpnTunnel"`
```
aquamatthias, line 307: This is not a property of a resource - please remove it. Could you make it a property in api_spec?

plugins/gcp/resoto_plugin_gcp/resources/base.py
```diff
@@ -301,14 +302,22 @@ class GcpResource(BaseResource):
...
+
+    @property
+    def _set_label_identifier(self) -> str:
+        return "resource"
```
aquamatthias, line 311: Isn't that also static information? Can we move it into ApiSpec?

plugins/gcp/resoto_plugin_gcp/resources/base.py
anjafr: Yes and no. `GcpResource`s should have more than one "api spec". One for collecting, using list or aggregateList actions, one for `get`ting an individual resource, one for deletion and one for updating tags. Currently we just say that the one api spec is the collection use case. In this PR so far I went down the route to modify (guess, really) the generated api spec to comply with the `setLabels` and `get` use case, where ideally we should have those generated with `model_gen.py` as well.
Mixing the multiple use cases into the one api spec just feels wrong, like having the `get` specific identifier as a property on the otherwise `aggregateList` api spec. What do you think?

anjafr: Heres an update that moves the multiple changes into a the ApiSpec class, sticking with collection use case as default, but extending the interface to get the `get` and `setLabels` specific specs from it as well: b4e21bdeba0869dfb3cffa41e79baab560b156eb

How do you feel about this approach?

aquamatthias: So yes - now it is static information attached to a resource class - that looks much better.
I also like your idea of creating specs on demand - I hope this works for all resource types?

anjafr: Looks like it so far. There's only one or two special snowflakes where I had to define deviating behaviour 👍🏻

anjafr:
Update: Can now tag and delete all resources. Buckets get emptied before deletion

lloesche:
Couple of issues after running first test collects:

1)
There are currently thousands of objects below the project. Ideally there should only be regions under a project.
In AWS we add the "global" region and put every global object there, let's do the same in GCP to not pollute the project space. The old collector does that too for routes, but it's a bad behaviour that we should not continue, esp. as we get more kinds.

2)
There is a resource kind "gcp_sql_flag" that shouldn't be a resource. When a DB is making use of certain flags we can add them as a nested property to that DB.
It makes no sense to have something like "cloudsql.enable_pglogical" or "commit_delay" which are essentially individual DB settings as a resource in the graph.

3)
I'm getting thousands of log lines like
```
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-highmem-48 (902048): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-highmem-64 (902064): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-highmem-8 (902008): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-highmem-80 (902080): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-highmem-96 (902096): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-standard-128 (901128): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-standard-16 (901016): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-standard-2 (901002): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
23-04-13 14:30:53|resotoworker-gcp|DEBUG|46803|gcp_testaccount-e2e  Unable to determine SKU(s) for gcp_machine_type n2-standard-32 (901032): ['services/6F81-5844-456A/skus/4AA7-6169-3F12 (4AA7-6169-3F12)', 'services/6F81-5844-456A/skus/AFF5-E97F-63FD (AFF5-E97F-63FD)', 'services/6F81-5844-456A/skus/CDAA-A05D-DE0B (CDAA-A05D-DE0B)', 'services/6F81-5844-456A/skus/F902-A6F0-7D1B (F902-A6F0-7D1B)']
```

4)
Upon completion I get a couple of hundred errors about duplicate checksums, like
```
23-04-13 14:30:58|resotoworker|ERROR|46677|resoto.worker-collector_1  Duplicate checksum 6eCNl9_k6jLEPJDx82x0XA for node gcp_node_type m1-node-96-1433 (919614330000) in graph!
23-04-13 14:30:58|resotoworker|ERROR|46677|resoto.worker-collector_1  Duplicate checksum LZoAjxS4SfK75ae8jE_NtA for node gcp_node_type m2-node-416-11776 (804416117760000) in graph!
23-04-13 14:30:58|resotoworker|ERROR|46677|resoto.worker-collector_1  Duplicate checksum 6eCNl9_k6jLEPJDx82x0XA for node gcp_node_type m1-node-96-1433 (919614330000) in graph!
```


I haven't yet compared the graph structure of the new vs. the old collector.

anjafr:
> 4. Upon completion I get a couple of hundred errors about duplicate checksums, like

Thats great feedback, thanks!

After looking into it for a few minutes it seems that we're correctly collecting `nodeType`s, but linking those to regions/zones doesn't work as expected at this point. The `undefined` zone name then leads to the duplicate object and checksum warnings.

```py
[nt.link.split('/')[8] for nt in collector.graph.search("name", "m1-node-96-1433") if nt.zone().name == "undefined"]
['us-east1-a', 'europe-west5-c', 'europe-west5-b', 'europe-west5-a', 'us-east7-b', 'us-east7-c']
```
Those are the zones that we get zonal nodeTypes for, but arent collected by resoto yet. Looking into the zone collection logic now - might be something with "hidden" or deactivated zones, based on project (i.e. some projects can't use `us-east-1a`)

anjafr:
> Those are the zones that we get zonal nodeTypes for, but arent collected by resoto yet. Looking into the zone collection logic now - might be something with "hidden" or deactivated zones, based on project (i.e. some projects can't use `us-east-1a`)

Those zones are indeed simply not there for all projects. By removing `nodeType` nodes which aren't used by more specific resources, we should get rid of this problem, so I've added them to the node-removal that we already have in place for `machineType`s and so on

anjafr:
> 1. There are currently thousands of objects below the project. Ideally there should only be regions under a project.
In AWS we add the "global" region and put every global object there, let's do the same in GCP to not pollute the project space. The old collector does that too for routes, but it's a bad behaviour that we should not continue, esp. as we get more kinds.

What kind of resources are those? Looking at a list of non-regional resources that I compiled back in March thats likely:
`gcp_firewall, gcp_interconnect_location, gcp_network, gcp_service, gcp_sku, gcp_route, gcp_sql_flag`, or?


* For `gcp_service` and  `gcp_sku` we should even consider pulling them upwards under the `Gcp` cloud resource, as those are not project specific but global and shared across all Gcp customers and projects.
* `gcp_sql_flag` are no longer collected as per another commit earleir today.
* `gcp_firewall, gcp_interconnect_location, gcp_network and gcp_route` - Looking into linking them to a new global region now.

Note:
There is also `firewallPolicies` - which we arent collecting yet at all - that is not project specific, but one level up on a GCP organisation. I'd consider this one out of scope for this PR.

anjafr:
With 3b6394a15e353e67d02f7ec5a246ef3119113403, all non-zonal and non-regional resources are connected to a new (fake) `global` region resource, instead of the project.

anjafr:
@lloesche looking for more feedback as soon as you're able to re-run the collection on those projects please 😄

lloesche:
Nice, I just tested and looks pretty good.

The region `global` is not fake as far as Google is concerned. For many services it's part of the identifier. E.g. `projects/{{project}}/global/networks/{{name}}` for networks.

> There is also firewallPolicies - which we arent collecting yet at all - that is not project specific, but one level up on a GCP organisation. I'd consider this one out of scope for this PR.

Are they related to the `gcp_firewall` resource that we are collecting? If yes, we should add them if it's not a lot of effort. If they are unrelated then I agree, let's skip them for now.


I still got a few duplicate checksums but a lot fewer than previously.

The global region seems to exist multiple times in the graph.
```
23-04-20 15:35:50|resotoworker|ERROR|3768341|resoto.worker-collector_1  Duplicate checksum Zt5JR3Fb0eFCJkwdnng-XQ for node gcp_region global in graph!
```
We must make sure that global is only being instantiated once per project. And that object instance then reused everywhere. If we create the same GcpRegion twice, even if it has the same id it's handled as a unique new object.

And then gcp_accelerator_type exists multiple times with the same checksum.
```
23-04-20 15:35:52|resotoworker|ERROR|3768341|resoto.worker-collector_1  Duplicate checksum ZQjsfBUlSEQABTmFWJ369w for node gcp_accelerator_type nvidia-tesla-p4 (10010) in graph!
23-04-20 15:35:52|resotoworker|ERROR|3768341|resoto.worker-collector_1  Duplicate checksum QbXgOKqIIqr6VSWbgdVJYw for node gcp_accelerator_type nvidia-tesla-p4-vws (10012) in graph!
```

Regarding GcpSqlFlag, we do want to collect them, but not as distinct resources. Instead they are database flags for Cloud SQL DB instances. They are attributes of the DB, sort of like the color of your car is an attribute of the car and not a distinct resource. Like, you don't go buy a car and the color red. You buy a red car, if that makes sense.
In the same way we should collect the flags associated with a Cloud SQL instance.

I still have 421 times "Unable to determine SKU" in a single project. Is that expected?

I noticed that the `id` field sometimes contains the GCP internal id. In the past we said that in our unified data model `id` contains the string that you require to talk to the API. Whereas `name` contains whatever name the user chose.

```
reported:
  id: '2543460420143619458'
  tags: {}
  name: konvoy-e2e-bastion-1681957679-bastion
  ctime: '2023-04-20T02:29:02Z'
  instance_cores: 4
  instance_memory: 16
  instance_type: n2-standard-4
  instance_status: running
```

In this case, the internal Google ID is practically useless for the user. And what's currently in `name` is the id that you use to talk to the Google APIs.

To be consistent we should have ID contain the ID that the API expects and maybe have a `google_id` or something like that where we can map the google internal ID.

@aquamatthias @anjafr thoughts?

anjafr:
> Regarding GcpSqlFlag, we do want to collect them, but not as distinct resources. Instead they are database flags for Cloud SQL DB instances. They are attributes of the DB, sort of like the color of your car is an attribute of the car and not a distinct resource. Like, you don't go buy a car and the color red. You buy a red car, if that makes sense.
> In the same way we should collect the flags associated with a Cloud SQL instance.

I had looked into this previously and surprisingly, there is no way to link the flag to a database instance, from either direction.

anjafr:
> > There is also firewallPolicies - which we arent collecting yet at all - that is not project specific, but one level up on a GCP organisation. I'd consider this one out of scope for this PR.
> 
> Are they related to the `gcp_firewall` resource that we are collecting? If yes, we should add them if it's not a lot of effort. If they are unrelated then I agree, let's skip them for now.

Those policies can't be collected by the current scope of the project collector, because they're on org level. So yes, that would be a bigger  chunk of work and out of scope for parity with the existing collector, that also doesn't do any of that. See more details [in our trello](https://trello.com/c/3SJE1AHD). Just mentioned it for sake of completeness 😄

aquamatthias:
> > Regarding GcpSqlFlag, we do want to collect them, but not as distinct resources. Instead they are database flags for Cloud SQL DB instances. They are attributes of the DB, sort of like the color of your car is an attribute of the car and not a distinct resource. Like, you don't go buy a car and the color red. You buy a red car, if that makes sense.
> > In the same way we should collect the flags associated with a Cloud SQL instance.
> 
> I had looked into this previously and surprisingly, there is no way to link the flag to a database instance, from either direction.

Isn't that simply a new property on the database instance?

anjafr:
> > > Regarding GcpSqlFlag, we do want to collect them, but not as distinct resources. Instead they are database flags for Cloud SQL DB instances. They are attributes of the DB, sort of like the color of your car is an attribute of the car and not a distinct resource. Like, you don't go buy a car and the color red. You buy a red car, if that makes sense.
> > > In the same way we should collect the flags associated with a Cloud SQL instance.
> > 
> > 
> > I had looked into this previously and surprisingly, there is no way to link the flag to a database instance, from either direction.
> 
> Isn't that simply a new property on the database instance?

Sql Flags are referencing [database versions, not instances](https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/flags#Flag) - so, no 🤷🏻‍♀️

anjafr:
> The global region seems to exist multiple times in the graph.

fixed

anjafr:
> And then gcp_accelerator_type exists multiple times with the same checksum.

fixed

anjafr:
> I still have 421 times "Unable to determine SKU" in a single project. Is that expected?

This can only come from GcpDiskTypes - but I can't reproduce on my project

lloesche:
> > Isn't that simply a new property on the database instance?
> 
> Sql Flags are referencing [database versions, not instances](https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1/flags#Flag) - so, no 🤷🏻‍♀️
> 
> ETA: I want to point out that even in my small project, without any database instances, we collected over 500 sql flags (see first PR comment). Those are all just some sort of default flag, it's just clutter, really.

Yes, SQL flags are essentially just a dictionary on a DB instance with the flag itself as key and flag value as value. We could even flatten it down to be a string as key and union of base types as value. When you create a new DB instance you  configure its flags and flag values like so:
image

lloesche:
I just ran a collect against our own account and created a MySQL DB instance called `testdb` with some random SQL flag `flush_time` on it. First of all, I see the flags when I use `gcloud sql instances describe testdb` on the CLI:
```
  dataDiskSizeGb: '100'
  dataDiskType: PD_SSD
  databaseFlags:
  - name: flush_time
    value: '3600'
  deletionProtectionEnabled: true
  insightsConfig: {}
```
So I would think the API returns them? Or do we already collect them?

I was trying to find out the later but the collector died with the following message:
```
23-04-25 13:33:18|resotoworker-gcp|ERROR|94144|gcp_inbound-axon-320811  [gcp:inbound-axon-320811] Failed to collect project: Could not match the object of type "dict" to any type of the Union: GcpSqlSettings, NoneType
```

You can replicate this by creating a DB instance and then testing the collector code.

lloesche:
> @lloesche hang on - it's just hidden in plain sight. Our model of `GcpDatabaseInstance` has a property `settings`, which in turn contains `database_flags` - so I think we're good, actually 😅

Yes I think so. That's what I half suspected and wanted to test when the collector had some issue mapping the value it encountered from the API.

anjafr:
> I noticed that the `id` field sometimes contains the GCP internal id. In the past we said that in our unified data model `id` contains the string that you require to talk to the API. Whereas `name` contains whatever name the user chose.

fixed (I think)
I looked at where we individually `get` resources, and in all cases they use the human readable `name`, not the `id`. So I went ahead and made that the default. Nothing broke 🤞🏻

plugins/gcp/resoto_plugin_gcp/resources/sqladmin.py
```diff
@@ -578,7 +578,7 @@ class GcpSqlDatabaseInstance(GcpResource):
     secondary_gce_zone: Optional[str] = field(default=None)
     server_ca_cert: Optional[GcpSqlSslCert] = field(default=None)
     service_account_email_address: Optional[str] = field(default=None)
-    settings: Optional[GcpSqlSettings] = field(default=None)
+    settings: GcpSqlSettings = field(default=None)
```
lloesche: Is this correct? If the field is no longer optional, then the default can't be None.

anjafr: It is still optional, that was just an experimental change. But I rolled it back in https://github.com/someengineering/resoto/pull/1528/commits/ae6308144b5725f05ffa514d93aaffd3b63ede67 🤔

aquamatthias (commented):
In order to merge this PR - the old collector needs to be removed.

lloesche:
Some more collect issues:
1) lot's of None's - the problematic one is `id` but maybe others as well
```
23-05-03 13:59:39|resotoworker-gcp|DEBUG|3659678|gcp_braided-flow-431_1  Can not deserialize json into class GcpSqlUser: {'id': None, 'tags': {}, 'name': None, 'ctime': None, 'description': None, 'link': None, 'label_fingerprint': None, 'deprecation_status': None, 'dual_password_type': None, 'etag': '2e23234bc26e1b2ac274ccd5a0bd60d5ae5a92bf4ec244674073327d2a23a69e', 'host': 'cloudsqlproxy~%', 'instance': 'review-board', 'password': None, 'password_policy': None, 'project': 'braided-flow-431', 'sqlserver_user_details': None, 'type': None}. Error: NoneType cannot be deserialized into str
```

2) Some cyclic dependency
```
23-05-03 13:50:50|resotoworker|ERROR|3606164|collector_gcp  Graph of plugin gcp is not acyclic - ignoring plugin results. Cycle (<built-in method join of str object at 0x7f42629ee4f0>, ['EdgeType.default: gcp_target_vpn_gateway mcd-vpn-->gcp_forwarding_rule mcd-vpn-rule-udp500', 'EdgeType.default: gcp_forwarding_rule mcd-vpn-rule-udp500-->gcp_target_vpn_gateway mcd-vpn'])
23-05-03 13:50:50|resotoworker|ERROR|3605756|resoto.worker-collector_1  Skipping invalid collector graph <class 'NoneType'>
```

3) If someone doesn't use K8S (or any other API) we shouldn't abort the entire project.
```
[gcp:eng-mesosphere-dcos] Failed to collect project: <HttpError 403 when requesting https://container.googleapis.com/v1/projects/eng-mesosphere-dcos/locations/-/operations?alt=json returned "Kubernetes Engine API has not been used in project 958846831018 before or it is disabled
```

lloesche:
Some more collect issues:
1) lot's of None's - the problematic one is `id` but maybe others as well
```
23-05-03 13:59:39|resotoworker-gcp|DEBUG|3659678|gcp_braided-flow-431_1  Can not deserialize json into class GcpSqlUser: {'id': None, 'tags': {}, 'name': None, 'ctime': None, 'description': None, 'link': None, 'label_fingerprint': None, 'deprecation_status': None, 'dual_password_type': None, 'etag': '2e23234bc26e1b2ac274ccd5a0bd60d5ae5a92bf4ec244674073327d2a23a69e', 'host': 'cloudsqlproxy~%', 'instance': 'review-board', 'password': None, 'password_policy': None, 'project': 'braided-flow-431', 'sqlserver_user_details': None, 'type': None}. Error: NoneType cannot be deserialized into str
```

2) Some cyclic dependency
```
23-05-03 13:50:50|resotoworker|ERROR|3606164|collector_gcp  Graph of plugin gcp is not acyclic - ignoring plugin results. Cycle (<built-in method join of str object at 0x7f42629ee4f0>, ['EdgeType.default: gcp_target_vpn_gateway mcd-vpn-->gcp_forwarding_rule mcd-vpn-rule-udp500', 'EdgeType.default: gcp_forwarding_rule mcd-vpn-rule-udp500-->gcp_target_vpn_gateway mcd-vpn'])
23-05-03 13:50:50|resotoworker|ERROR|3605756|resoto.worker-collector_1  Skipping invalid collector graph <class 'NoneType'>
```

3) If someone doesn't use K8S (or any other API) we shouldn't abort the entire project.
```
[gcp:braided-flow-431_1] Failed to collect project: <HttpError 403 when requesting https://container.googleapis.com/v1/projects/braided-flow-431_1/locations/-/operations?alt=json returned "Kubernetes Engine API has not been used in project 958846831018 before or it is disabled
```

anjafr:
> 2\. Some cyclic dependency

resolved

lloesche:
> > 2. Some cyclic dependency
> 
> resolved

Now the dependency that resulted in the cyclic dependency is just deleted. Meaning it was wrong in the first place? Or is it missing now?

anjafr:
> Now the dependency that resulted in the cyclic dependency is just deleted. Meaning it was wrong in the first place? Or is it missing now?

it was defined from both ends, I deleted one of the definitions only 👍🏻

lloesche (approved):

lloesche:
I had this now run against D2iQ many times. With the latest changes it no longer throws any errors other than the "API not active" ones. We already talked about how we'll change the behaviour to skip resources for disabled APIs instead of skipping the project. But we will do so in a separate PR. I also have a gut feeling that we'll run into more issues on the delete side of things, but merging this into `main` will allow our GCP users to test the `:edge` image before we make it into a numbered release. We won't cut a release this Friday so we'll have 1.5 weeks of additional edge case testing.
So from my PoV this is good to merge.

PR #1617, opened 2023-05-30
[plugins/k8s][feat] Store LB backends when collecting ingress

meln1k:
# Description

# To-Dos

# Issues Fixed

- Fixes #XXXX

# Code of Conduct

### DIFF
--- a/plugins/k8s/resoto_plugin_k8s/resources.py
+++ b/plugins/k8s/resoto_plugin_k8s/resources.py
@@ -2297,7 +2297,7 @@ class KubernetesIngress(KubernetesResource, BaseLoadBalancer):
 
                 for key, value in selector.items():
                     for pod in pods_by_labels.get((key, value), []):
-                        resolved_backends.add(pod.name)
+                        resolved_backends.add(pod.name or pod.id)
 
         self.backends = list(resolved_backends)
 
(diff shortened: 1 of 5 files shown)


aquamatthias (dismissed):

plugins/k8s/resoto_plugin_k8s/resources.py
```diff
@@ -2237,6 +2237,7 @@ class KubernetesIngress(KubernetesResource, BaseLoadBalancer):
         "public_ip_address": S("status", "loadBalancer", "ingress", default=[])[0]["ip"],
         # take the public ip of the first load balancer
         "ingress_spec": S("spec") >> Bend(KubernetesIngressSpec.mapping),
+        "backends": S("spec") >> S("rules", default=[]) >> ForallBend(S("host")),
```
lloesche, line 2240: I'm afraid this is too simplistic. In AWS and GCP backends[] holds a list of strings with the IDs of the backend servers. So e.g. with ALBs it resolves all of the ALB Target Groups and finds the backend instances.
Translated to K8S it should hold the list of pods that traffic ends up on. To do that you not only need to evaluate the `rules[*].http.paths[*].backend.service.name` but also `spec.defaultBackend.service.name` (if present) and then check the `spec.selector` of each service. Then match `metadata.labels` of each pod against the selector and if it matches add the pod to the list of backends. That way you know if a loadbalancer is abandoned or not.

meln1k: Thanks for the insights!

PR #1654, opened 2023-06-12
[resoto][chore] Remove plugins that are now InfraApps

aquamatthias:
# Code of Conduct


TheCatLady:
Several of our how-to guides (e.g., https://resoto.com/docs/how-to-guides/cleanup/clean-up-expired-resources) use plugins. We should update these prior to removing these plugins.

There's also a plugins section we can just remove (https://resoto.com/docs/reference/components/plugins), but we should make sure there are appropriate places to redirect each of these -- perhaps to the top-level module pages until we have documentation of the apps? 🤔  Or, do we just rename the plugins section to "Infrastructure Apps"?

lloesche:
@TheCatLady the other concern is that plugins are not 1:1 compatible with Infra Apps. There is an infra app that's performing the same task, but the configuration has to be migrated by the users. That's why we'll keep the plugins until 3.7.

TheCatLady:
> @TheCatLady the other concern is that plugins are not 1:1 compatible with Infra Apps. For each plugin there is an infra app that's performing the same task, but the configuration has to be migrated by the users. That's why we'll keep the plugins until 3.7.

Sounds good. Prior to merging this, we should update the edge-versioned docs. Then, when 3.7 is released, we should update the 3.X docs.

aquamatthias:
Since this PR will not be merged in the next couple of days, I will close it. 
We can reopen it, once the time has come.

PR #1693, opened 2023-06-28
[plugins/aws][feat] Send EC2 cloudwath metrics while collecting

meln1k:
# Description

# To-Dos

# Issues Fixed

- Fixes #XXXX

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/configuration.py
+++ b/plugins/aws/resoto_plugin_aws/configuration.py
@@ -240,7 +240,7 @@ class AwsConfig:
             "If false, the error is logged and the resource is skipped."
         },
     )
-    collect_usage_metrics: bool = field(
+    collect_usage_metrics: Optional[bool] = field(
         default=True,
         metadata={"description": "Collect resource usage metrics via CloudWatch, enabled by default"},
     )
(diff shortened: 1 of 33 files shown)


plugins/aws/resoto_plugin_aws/configuration.py
```diff
@@ -240,6 +240,10 @@ class AwsConfig:
             "If false, the error is logged and the resource is skipped."
         },
     )
+    collect_usage_metrics: bool = field(
```
aquamatthias, line 243: Should be either optional or you need a migration. Existing installations do not have this property.

plugins/aws/resoto_plugin_aws/resource/apigateway.py
```diff
@@ -497,6 +497,7 @@ def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) ->
                                     resource_instance.resource_methods[method] = gm
                         builder.add_node(resource_instance, resource)
                         builder.add_edge(api_instance, EdgeType.default, node=resource_instance)
+        return []
```
aquamatthias, line 500: What is the meaning? Everything that is collected via this method? only the resource type that is asked?
looks wrong - there are instances created. Also all other places that return [].

plugins/aws/resoto_plugin_aws/configuration.py
meln1k: won't default value be sufficient here?

PR #1711, opened 2023-07-06
[resoto][feat] Send individual accounts to Resoto Core

lloesche:
# Description

Send individual accounts to Resoto Core to conserve memory.

This PR makes merge on account level the default and sends individual accounts to Resoto Core as soon as they are available, to free up the memory they use.

This removes all post_collect code since we no longer have a complete graph of all clouds in memory. The plugins aws_k8s and digitalocean_k8s need to be rewritten to use the post_collect event, search the core instead of in-memory graph and send deferred edges to the core. Likely the better approach anyways since this allows us to collect different clouds on different workers instead of relying on all K8S cluster being collected on the same worker their cloud is collected on.

# Code of Conduct

### DIFF
--- a/resotolib/resotolib/graph/__init__.py
+++ b/resotolib/resotolib/graph/__init__.py
@@ -402,6 +402,22 @@ def resource_classes_to_resotocore_model(classes: Set[Type[Any]], **kwargs: Any)
     return list(model.values())
 
 
+def export_model() -> List[Json]:
+    """Return the graph node dataclass model in resotocore format"""
+    model = dataclasses_to_resotocore_model({BaseResource}, aggregate_root=BaseResource)
+    for resource_model in model:
+        if resource_model.get("fqn") == "resource":
+            resource_model.get("properties", []).append(
+                {
+                    "name": "kind",
+                    "kind": "string",
+                    "required": True,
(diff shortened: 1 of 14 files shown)


resotolib/resotolib/graph/__init__.py
```diff
@@ -402,6 +402,22 @@ def resource_classes_to_resotocore_model(classes: Set[Type[Any]], **kwargs: Any)
     return list(model.values())
 
 
+def export_model() -> List[Json]:
+    """Return the graph node dataclass model in resotocore format"""
+    model = dataclasses_to_resotocore_model({BaseResource}, aggregate_root=BaseResource)
```
aquamatthias, line 407:
```suggestion
    model = resource_classes_to_resotocore_model({BaseResource}, aggregate_root=BaseResource)
```

Or maybe the 2 methods should be combined into one.

aquamatthias (dismissed):
Thanks!

resotolib/resotolib/graph/__init__.py
lloesche: Good idea, done.

aquamatthias (approved):

PR #1713, opened 2023-07-10
[k8s][feat] Add deferred edge linking

lloesche:
# Description

Add deferred edge linking

# Code of Conduct


plugins/k8s/resoto_plugin_k8s/__init__.py
```diff
@@ -108,6 +109,7 @@ def collect_cluster(
             core_feedback.error(f"An unhandled error occurred while collecting {cluster_id}: {e}", log)
             raise
         else:
+            create_deferred_edges(kc.graph)
```
aquamatthias, line 112: Can we do it inside the try block? Otherwise we will not see any error in case this method throws.

plugins/k8s/resoto_plugin_k8s/deferred_edges/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+def create_deferred_edges(graph: Graph) -> None:
+    if not isinstance(graph, Graph):
+        log.error(f"Expected type Graph, got {type(graph)}")
+        return
+    for node in graph.nodes:
+        if not isinstance(node, BaseResource):
+            log.warning(f"Node {node} is not a BaseResource")
+            continue
+        # AWS
```
aquamatthias: Is there a chance to detect if we are in EKS or DO? Would be great to only add deferred edges if required.

aquamatthias (commented):
Nice 2 little questions/remarks.

plugins/k8s/resoto_plugin_k8s/deferred_edges/__init__.py
lloesche: Maybe, but I would make that an improvement in another PR. This PR just migrates the existing post_collect logic without changing it.

I could add a config.resotoworker detection to see if AWS is even activated and if not not run the AWS linking code. Though that seems error prone, because I could have one worker where AWS is running and another where K8S is running and would want them to link deferred edges properly.

aquamatthias (approved):
thanks!

PR #1738, opened 2023-07-19
[posthog][fix] Handle/Catch a possible conversion failure

1nv1:
…
# Description

Prevent converting any type/string that is neither a valid date string nor present.

# Code of Conduct

### DIFF
--- a/plugins/posthog/resoto_plugin_posthog/resources.py
+++ b/plugins/posthog/resoto_plugin_posthog/resources.py
@@ -106,8 +106,11 @@ class PosthogEvent(PosthogResource, BaseResource):
         )
 
 
-def convert_date(date_str: str) -> Optional[datetime]:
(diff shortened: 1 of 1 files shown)


plugins/posthog/resoto_plugin_posthog/resources.py
```diff
@@ -106,8 +106,11 @@ def new(data: Dict) -> BaseResource:
         )
 
 
-def convert_date(date_str: str) -> Optional[datetime]:
+def convert_date(date_str: str = '') -> Optional[datetime]:
     try:
-        return datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%S.%fZ")
+        if not isinstance(date_str, str):
```
aquamatthias, line 111: `date_str` is typed as str and then we check if the type is string.
This looks wrong. Do you want to prevent `None` or what other types can be expected?

PR #1778, opened 2023-09-19
[plugins/aws] Get account name from profile

lloesche:
# Description

Get account name from profile

# Code of Conduct

### DIFF
--- a/plugins/aws/resoto_plugin_aws/__init__.py
+++ b/plugins/aws/resoto_plugin_aws/__init__.py
@@ -454,7 +454,7 @@ def current_account_id_and_partition(profile: Optional[str] = None) -> Tuple[str
         raise botocore.exceptions.NoCredentialsError()
 
 
-def set_account_names(accounts: List[AwsAccount]) -> None:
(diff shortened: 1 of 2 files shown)


plugins/aws/resoto_plugin_aws/__init__.py
```diff
@@ -491,13 +491,36 @@ def set_name_from_org() -> bool:
...
+                account.name = account.profile
+                log.debug(f"Set name for {account.kdname} from profile")
+                return True
+            return False
+
+        # if we prefer the profile name and we have a profile
+        # we set the name from the profile and return immediately
+        if Config.aws.prefer_profile_as_account_name:
+            if Config.aws.scrape_org:
+                log.error("Setting account name from profile with scrape_org enabled is likely not what you want")
```
aquamatthias, line 505: All acounts in the org would have the same name?
If yes, we should ignore this setting in that case?
Logging the error will not propagate this information to the user. Consider using CoreFeedback.

aquamatthias (commented):
One question. Otherwise lgtm.

aquamatthias (approved):

PR #1812, opened 2023-11-01
[resotocore][feat] Add list table output

aquamatthias:
# Description

# To-Dos

# Code of Conduct

### DIFF
--- a/resotocore/resotocore/cli/command.py
+++ b/resotocore/resotocore/cli/command.py
@@ -2764,8 +2764,11 @@ class ListCommand(CLICommand, OutputTransformer):
             elif parsed.markdown:
                 return markdown_stream(in_stream)
             elif parsed.table:
-                dependencies = stream.call(partial(self.dependencies.model_handler.load_model, ctx.graph_name))
-                return stream.flatmap(dependencies, partial(table_stream, in_stream))
+
+                async def load_model() -> Model:
+                    return await self.dependencies.model_handler.load_model(ctx.graph_name)
+
+                return stream.flatmap(stream.call(load_model), partial(table_stream, in_stream))
             else:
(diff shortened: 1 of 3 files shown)


resotocore/resotocore/cli/command.py
```diff
@@ -2553,6 +2555,7 @@ def args_info(self) -> ArgsInfo:
         return [
             ArgInfo("--csv", help_text="format", option_group="format"),
             ArgInfo("--markdown", help_text="format", option_group="format"),
+            ArgInfo("--table", help_text="format", option_group="format"),
```
lloesche, line 2558: Should this be `list --json` Since list already implies that we're only outputting the list view, just that instead of markdown or key=value it's represented as json?

aquamatthias: I was starting like this as well and thought that `table` describes the output better 
```json
{columns=[....]}
{row={...}}
{row={...}}
{row={...}}
```

I can also download csv or markdown in json, ndjson, xml, etc.
```
$> echo "search is(instance) | list --csv" | http :8980/cli/execute "Content-Type: text/plain" 
HTTP/1.1 200 OK
Content-Type: application/json

[   "\"kind\",\"id\",\"name\",\"instance_memory\",\"age\",\"last_update\",\"cloud\",\"account\",\"region\",\"zone\"",
    "\"kubernetes_node\",\"0c6d10e4-d9d5-458d-a561-f6d7c1b9cd47\",\"fixsaas-ccx23-pool-jobs-worker1\",15.2413330078125,\"14d10h\",\"14d10h\",\"k8s\",\"dev\",\"default\",\"\""
]
```

What about `--json-table`?

lloesche: I like `--json-table`. It's very clear what you get with it.

PR #1832, opened 2023-11-21
[resotocore][feat] Enable count and total count for search queries

aquamatthias:
# Description

The search and cli/execute endpoint now return 2 additional headers:

```
Result-Count: 1
Total-Count: 61
```

One indicating the number of elements returned within this response, the other one shows the number of all available elements. There will be only a difference if a limit is applied to the search.

# Code of Conduct

### DIFF
--- a/resotocore/resotocore/cli/command.py
+++ b/resotocore/resotocore/cli/command.py
@@ -96,6 +96,7 @@ from resotocore.cli.model import (
     ArgInfo,
     EntityProvider,
     FilePath,
+    CLISourceContext,
 )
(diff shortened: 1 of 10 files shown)


lloesche:
All seems to work but I might have found a bug here:
```
> graph=d2iq search is(aws_ec2_instance) limit 0, 10 | count
total matched: 496
total unmatched: 0
```

When I use the `count` command, I want to count the number of elements that the previous command emitted.

lloesche (approved):
lgtm ty! 🚢 👍

PR #1852, opened 2023-12-07
[azure] [feat] Filter virtual machine sizes

1101-1:
# Description

## Objective:
Filter virtual machine sizes to retain only those associated with existing virtual machines.

# To-Dos
- [x] Filter virtual machine sizes to collect it correctly
- [x] Delete unnecessary network resource

# Code of Conduct

### DIFF
--- a/plugins/azure/resoto_plugin_azure/resource/base.py
+++ b/plugins/azure/resoto_plugin_azure/resource/base.py
@@ -4,7 +4,7 @@ import logging
 from concurrent.futures import Future
 from datetime import datetime
 from threading import Lock
-from typing import Any, ClassVar, Dict, Optional, TypeVar, List, Tuple, Type, Callable, Union, cast
+from typing import Any, ClassVar, Dict, Optional, TypeVar, List, Type, Callable, cast
 
 from attr import define, field
 from azure.core.utils import CaseInsensitiveDict
@@ -133,40 +133,6 @@ class AzureResource(BaseResource):
     def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
         # Default behavior: add resource to the namespace
         pass
-
-    def fetch_resources(
-        self,
-        builder: GraphBuilder,
-        service: str,
-        api_version: str,
-        path: str,
-        path_parameters: List[str],
-        query_parameters: List[str],
-        first_property: Callable[[Json], Union[List[str], str]],
-        second_property: Callable[[Json], Union[List[str], str]],
-    ) -> List[Tuple[Union[str, List[str]], Union[str, List[str]]]]:
-        """
-        Fetch additional resources from the Azure API.
-        Can used for further connection using the connect_in_graph method.
-
-        Parameters:
-        - first_property: Compared property.
-        - second_property: Binding property | Compared property.
-
-        Returns:
-        List[Tuple[Union[str, List[str]], Union[str, List[str]]]]: A list of tuples containing information to compare and connect the retrieved resources.
-        """
-        resources_api_spec = AzureApiSpec(
-            service=service,
-            version=api_version,
-            path=path,
-            path_parameters=path_parameters,
-            query_parameters=query_parameters,
-            access_path="value",
-            expect_array=True,
-        )
-
-        return [(first_property(r), second_property(r)) for r in builder.client.list(resources_api_spec)]
 
     @classmethod
     def collect_resources(
--- a/plugins/azure/resoto_plugin_azure/resource/base.py
+++ b/plugins/azure/resoto_plugin_azure/resource/base.py
@@ -134,32 +134,8 @@ class AzureResource(BaseResource):
         # Default behavior: add resource to the namespace
         pass
 
-    def filter_node(self, compare_resources: Optional[List[Tuple[str, str]]]) -> bool:
-        """
-        Determine whether the current instance meets exclusion criteria based on a list of resource tuples for comparison.
-
-        Parameters:
-        - compare_resources (Optional[List[Tuple[str, str]]]): A list of resource tuples for comparison.
-
-        Returns:
-        bool: True if the instance meets exclusion criteria; False if it does not.
-
-        Default: return False. It means that, by default, the node does not need to be excluded.
-        """
-        return False
-
-    @classmethod
-    def collect_resources_for_comparison(
-        cls: Type[AzureResourceType], builder: GraphBuilder
-    ) -> Optional[List[Tuple[str, str]]]:
-        """
-        Perform any pre-collection steps and return a list of items for comparison if needed.
-        """
-        return None
-
-    @classmethod
     def fetch_resources(
-        cls: Type[AzureResourceType],
+        self,
         builder: GraphBuilder,
         service: str,
         api_version: str,
--- a/plugins/azure/resoto_plugin_azure/resource/base.py
+++ b/plugins/azure/resoto_plugin_azure/resource/base.py
@@ -134,6 +134,16 @@ class AzureResource(BaseResource):
         # Default behavior: add resource to the namespace
         pass
 
+    def should_add_to_node(self, builder: GraphBuilder, source: Json, pre_items: Optional[List[str]] = None) -> bool:
+        """
+        Checks whether this instance should be added to a graph node.
+
+        Returns:
+        bool: True if the instance should be added to the graph node, False otherwise.
+        Default: return True.
+        """
+        return True
+
     def fetch_resources(
         self,
         builder: GraphBuilder,
@@ -142,11 +152,16 @@ class AzureResource(BaseResource):
         path: str,
         path_parameters: List[str],
         query_parameters: List[str],
-        compared_property: Callable[[Json], Union[List[str], str]],
-        binding_property: Callable[[Json], Union[List[str], str]],
+        first_property: Callable[[Json], Union[List[str], str]],
+        second_property: Callable[[Json], Union[List[str], str]],
     ) -> List[Tuple[Union[str, List[str]], Union[str, List[str]]]]:
         """
-        Fetch additional resources from the Azure API for further connection using the connect_in_graph method.
+        Fetch additional resources from the Azure API.
+        Can used for further connection using the connect_in_graph method.
+
+        Parameters:
+        - first_property: Compared property.
+        - second_property: Binding property | Compared property.
 
         Returns:
         List[Tuple[Union[str, List[str]], str]]: A list of tuples containing information to compare and connect the retrieved resources.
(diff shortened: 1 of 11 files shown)


plugins/azure/resoto_plugin_azure/resource/compute.py
```diff
@@ -2612,6 +2613,27 @@ class AzureVirtualMachine(AzureResource, BaseInstance):
...
+                access_path="value",
+                expect_array=True,
+            )
+            hardware_profile_name = self.instance_type if self.instance_type else ""
+            location = self.location if self.location else ""
+            items = [
+                r for r in graph_builder.client.list(api_spec, location=location) if r["name"] == hardware_profile_name
+            ]
+            AzureVirtualMachineSize.collect(items, graph_builder)
+
```
aquamatthias, line 2635: This will be executed for every virtual machine. 
If I have 20K instances of one size, we would have 20K same virtual machine sizes.
It is also very inefficient to fetch.
I would recommend to fetch all virtual machine sizes and than remove all sizes that are not used.

plugins/azure/resoto_plugin_azure/resource/base.py
```diff
@@ -187,11 +216,13 @@ def collect(cls: Type[AzureResourceType], raw: List[Json], builder: GraphBuilder
             # map from api
             instance = cls.from_api(js)
             instance.pre_process(builder, js)
-            # add to graph
-            if (added := builder.add_node(instance, js)) is not None:
-                # post process
-                added.post_process(builder, js)
-                result.append(added)
+            # Check if there is a filter condition specified before adding the instance to the graph node.
+            if instance.should_add_to_node(builder, js, pre_items):
```
1101-1, line 220: I think that validate virtual machine sizes before we add to node them is better way. Because If we will iterate our nodes after collecting them, we would do more actions and spent more time.

aquamatthias: This only works if all virtual machines are collected beforehand.
I see you introduced `pre_collect`.  This might be a valid solution if we have one or two dependencies. But we actually have an arbitrary number of dependencies. Solving them by collecting them in different steps will make the code hard to understand and the collection time long (since only some things can be gathered in parallel). 

Fetching resources, adding them to the graph, and removing them in case they are not connected feels like unnecessary work. But it allows us to fetch everything in parallel and reason about the state that is collected. 

I would opt to keep the two steps we have:
- collect all resources
- process and connect

plugins/azure/resoto_plugin_azure/resource/compute.py
```diff
@@ -3231,6 +3233,10 @@ class AzureVirtualMachineSize(AzureResource, BaseInstanceType):
     number_of_cores: Optional[int] = field(default=None, metadata={'description': 'The number of cores supported by the virtual machine size. For constrained vcpu capable vm sizes, this number represents the total vcpus of quota that the vm uses. For accurate vcpu count, please refer to https://docs. Microsoft. Com/azure/virtual-machines/constrained-vcpu or https://docs. Microsoft. Com/rest/api/compute/resourceskus/list.'})  # fmt: skip
     os_disk_size_in_mb: Optional[int] = field(default=None, metadata={'description': 'The os disk size, in mb, allowed by the virtual machine size.'})  # fmt: skip
     resource_disk_size_in_mb: Optional[int] = field(default=None, metadata={'description': 'The resource disk size, in mb, allowed by the virtual machine size.'})  # fmt: skip
+    location: Optional[str] = field(default=None, metadata={"description": "Resource location."})
+
+    def pre_process(self, graph_builder: GraphBuilder, source: Json) -> None:
+        self.location = graph_builder.location.name if graph_builder.location else ""
```
aquamatthias, line 3239: All location-dependent resources use the same pattern: they are loaded via a GraphBuilder with location.
Please see GraphBuilder.add_node - there is already handling for defining the region in this case. Maybe this is already sufficient?

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -2858,8 +2834,8 @@ def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
             path="/subscriptions/{subscriptionId}/providers/Microsoft.Network/networkInterfaces",
             path_parameters=["subscriptionId"],
             query_parameters=["api-version"],
-            compared_property=lambda r: r["id"],
-            binding_property=binding_property,
+            first_property=lambda r: r["id"],
+            second_property=second_property,
```
aquamatthias, line 2838: This call is not required - it fetches all network interfaces, which have been already fetched and are available in the graph.

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -4266,8 +4231,8 @@ def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
             path="/subscriptions/{subscriptionId}/providers/Microsoft.Network/networkInterfaces",
             path_parameters=["subscriptionId"],
             query_parameters=["api-version"],
-            compared_property=compared_property,
-            binding_property=lambda r: r["properties"]["virtualMachine"]["id"]
+            first_property=first_property,
+            second_property=lambda r: r["properties"]["virtualMachine"]["id"]
             if "properties" in r and "virtualMachine" in r["properties"]
             else "",
         )
```
aquamatthias, line 4238: This call is not required - it fetches all network interfaces, which have been already fetched and are available in the graph.

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -5008,8 +4966,8 @@ def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
             path="/subscriptions/{subscriptionId}/providers/Microsoft.Network/networkInterfaces",
             path_parameters=["subscriptionId"],
             query_parameters=["api-version"],
-            compared_property=compared_property,
-            binding_property=binding_property,
+            first_property=first_property,
+            second_property=second_property,
         )
```
aquamatthias, line 4971: This call is not required - it fetches all network interfaces, which have been already fetched and are available in the graph.

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -3844,8 +3809,8 @@ def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
             path="/subscriptions/{subscriptionId}/providers/Microsoft.Network/virtualNetworks",
             path_parameters=["subscriptionId"],
             query_parameters=["api-version"],
-            compared_property=lambda r: r["properties"]["addressSpace"]["addressPrefixes"],
-            binding_property=lambda r: r["id"],
+            first_property=lambda r: r["properties"]["addressSpace"]["addressPrefixes"],
+            second_property=lambda r: r["id"],
         )
```
aquamatthias, line 3814: This call is not required - it fetches all virtual networks, which have been already fetched and are available in the graph.

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -4516,14 +4474,14 @@ class AzureNetworkWatcher(AzureResource):
...
             service="network",
             api_version="2023-05-01",
             path="/subscriptions/{subscriptionId}/providers/Microsoft.Network/virtualNetworks",
             path_parameters=["subscriptionId"],
             query_parameters=["api-version"],
-            compared_property=lambda r: r["location"],
-            binding_property=lambda r: r["id"],
+            first_property=lambda r: r["location"],
+            second_property=lambda r: r["id"],
         )
```
aquamatthias, line 4485: This call is not required - it fetches all virtual networks, which have been already fetched and are available in the graph.

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -3148,8 +3123,8 @@ def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
             path="/subscriptions/{subscriptionId}/providers/Microsoft.Network/ExpressRoutePortsLocations",
             path_parameters=["subscriptionId"],
             query_parameters=["api-version"],
-            compared_property=lambda r: r["name"],
-            binding_property=lambda r: r["id"],
+            first_property=lambda r: r["name"],
+            second_property=lambda r: r["id"],
```
aquamatthias, line 3127: This call is not required - it fetches all express route locations, which have been already fetched and are available in the graph.

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -4426,22 +4390,16 @@ def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
             path="/subscriptions/{subscriptionId}/providers/Microsoft.Network/networkVirtualApplianceSkus",
             path_parameters=["subscriptionId"],
             query_parameters=["api-version"],
-            compared_property=lambda r: r["properties"]["vendor"],
-            binding_property=lambda r: r["name"],
+            first_property=lambda r: r["properties"]["vendor"],
+            second_property=lambda r: r["name"],
         )
```
aquamatthias, line 4395: This call is not required - it fetches all virtual networks, which have been already fetched and are available in the graph.

aquamatthias (changes requested):
Filtering is fine - thanks. One small comment on region handling.

I also stumbled upon `fetch_resouces`. All usages of this method are not required, since the related data is already available in the graph. Please remove this method and rework the usage.

plugins/azure/resoto_plugin_azure/resource/compute.py
1101-1: We need to know location to connect `AzureVirtualMachineSize` between `AzureVirtualMachine`. VirtualMachineSize by its own does not have location value.

And in GraphBuilder.add_node `add_node` method does not apply location on resource for further comparison.

plugins/azure/resoto_plugin_azure/resource/network.py
```diff
@@ -2846,21 +2822,18 @@ class AzureDscpConfiguration(AzureResource):
     source_port_ranges: Optional[List[AzureQosPortRange]] = field(default=None, metadata={'description': 'Sources port ranges.'})  # fmt: skip
 
     def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
-        binding_property: Callable[[Json], List[str]] = lambda r: [  # pylint: disable=unnecessary-lambda-assignment
-            ip_config["properties"]["subnet"]["id"]
-            for ip_config in r.get("properties", {}).get("ipConfigurations", [])
-            if "subnet" in ip_config.get("properties", {})
+        first_property: Callable[[AzureNetworkInterface], str] = lambda n: n.id or ""
+
+        second_property: Callable[[AzureNetworkInterface], List[str]] = lambda n: [
```
aquamatthias, line 2827: instead of calling those lambdas first and second, consider to use what they give back: `network_id` and `subnet_id` .
same applies to all other first and second properties

plugins/azure/resoto_plugin_azure/resource/compute.py
aquamatthias: It does. After you added the vmSize to the graph, the location is available via `region`.
No objection to introduce it as separate property.

aquamatthias (commented):
One request to rename first and second lambdas to something more intuitive. Otherwise lgtm.

aquamatthias (approved):

PR #1897, opened 2024-02-02
[resotolib][feat] Set replace on org root for account merge

lloesche:
# Description

Set replace on org root for account merge

# Code of Conduct


aquamatthias (changes requested):
Replacing at the level of org root together with one account would result in deleting all other accounts under this parent node.  That's probably not what we want.

PR #1993, opened 2024-03-14
[azure] [feat]: Add improved request/error limit handling

1101-1:
# Description

# To-Dos

# Code of Conduct


plugins/azure/fix_plugin_azure/azure_client.py
```diff
@@ -78,23 +109,20 @@ def __init__(
...
+            return False  # Resource not found to delete
         except HttpResponseError as e:
-            if e.error and e.error.code == "ResourceNotFoundError":
-                return False  # Resource not found to delete
-            else:
-                raise e
-
+            raise e
+        except Exception as e:
+            raise e
```
aquamatthias, line 125: useless: catching and raising does mot have any effect.

plugins/azure/fix_plugin_azure/azure_client.py
```diff
@@ -21,6 +24,26 @@
 from fix_plugin_azure.config import AzureCredentials
 from fixlib.types import Json
 
+log = logging.getLogger("fix.plugins.azure")
+
+
+def is_retryable_exception(e: Exception) -> bool:
+    # Placed there to avoid circular import issues
+    from fix_plugin_azure.resource.metrics import MetricRequestError
```
aquamatthias, line 32: Id it is a http client error, it should be defined in this module

plugins/azure/fix_plugin_azure/azure_client.py
```diff
@@ -128,16 +156,45 @@ def _update_or_delete_tag(self, tag_name: str, tag_value: str, resource_id: str,
...
         except HttpResponseError as e:
-            if e.error and e.error.code == "ResourceNotFoundError":
-                return False  # Resource not found
-            elif e.error and e.error.code == "ResourceExistsError":
-                return False  # Tag for update/delete does not exist
-            else:
-                raise e
+            raise e
+        except Exception as e:
+            raise e
```
aquamatthias, line 166: same here

plugins/azure/fix_plugin_azure/azure_client.py
```diff
@@ -128,16 +156,45 @@ def _update_or_delete_tag(self, tag_name: str, tag_value: str, resource_id: str,
...
 
         return True
 
+    @retry(  # type: ignore
+        stop_max_attempt_number=3,  # max 3 attempts
```
aquamatthias, line 171: There are 10 in AWS, and I would also use it here.
Fix is trying hard to get the data - otherwise, we would see an incomplete state.

plugins/azure/fix_plugin_azure/azure_client.py
```diff
@@ -128,16 +156,45 @@ def _update_or_delete_tag(self, tag_name: str, tag_value: str, resource_id: str,
...
+        except ClientAuthenticationError as e:
+            log.error(f"[Azure] Call to Azure API is not authorized!: {e}")
+            return None
+        except HttpResponseError as e:
+            if e.error:
+                if e.error.code == "NoRegisteredProviderFound":
+                    log.warning("[Azure] API not available in this region")
+                    return None
+                elif e.error.code == "BadRequest" and spec.service == "metric":
+                    raise MetricRequestError from e
```
aquamatthias, line 191: Can we narrow it down even further? What does this error mean?

plugins/azure/fix_plugin_azure/resource/metrics.py
```diff
@@ -20,6 +20,11 @@
 log = logging.getLogger("fix.plugins.azure")
 
 
+class MetricRequestError(HttpResponseError):
```
aquamatthias, line 23: see earlier comment: move to client.

plugins/azure/test/conftest.py
```diff
@@ -52,6 +44,20 @@ def delete_resource_tag(self, tag_name: str, resource_id: str) -> bool:
     def update_resource_tag(self, tag_name: str, tag_value: str, resource_id: str) -> bool:
         return False
 
+    def get_with_retry(self, spec: AzureApiSpec, **kwargs: Any) -> List[Json]:
```
aquamatthias, line 47: I would say the name is misleading...
Why do we need to do this change?

aquamatthias (commented):
Added a couple of questions/remarks.

plugins/azure/fix_plugin_azure/azure_client.py
```diff
@@ -21,6 +23,28 @@
...
+
+class MetricRequestError(HttpResponseError):
+    """Error raised when there's an issue retrieving metric data from the Azure API."""
+
+
+def is_retryable_exception(e: Exception) -> bool:
+    # If we receive a metric request error, then repeat the request
+    if isinstance(e, MetricRequestError):
+        log.debug(f"Azure Metric request error occured, retrying: {e}")
+        return True
```
aquamatthias, line 37: An exception of this class is created in the exception handler and then catched here again.
Maybe it is simpler to match the original exception here?

plugins/azure/test/conftest.py
```diff
@@ -23,18 +23,10 @@
...
-                if spec.access_path:
-                    js = js[spec.access_path]
-            if spec.expect_array and isinstance(js, list):
-                return js
-            else:
-                return [js]
+        result = self.list_with_retry(spec, **kwargs)
+        if result is None:
+            return []
+        return result
```
aquamatthias, line 29: Just revert. there is no change and nothing is  retried.

plugins/azure/fix_plugin_azure/azure_client.py
1101-1: I use the `MetricRequestError` class to match the metric error exactly and retry the request for it. If we compare the original exception (`HttpResponseError`) with error_code as `BadRequest`, then for other resources (except metrics resources) having the same error code, we will try to get them again. And it doesn't suit us

plugins/azure/test/conftest.py
```diff
@@ -52,6 +51,9 @@ def delete_resource_tag(self, tag_name: str, resource_id: str) -> bool:
     def update_resource_tag(self, tag_name: str, tag_value: str, resource_id: str) -> bool:
         return False
 
+    def list_with_retry(self, spec: AzureApiSpec, **kwargs: Any) -> List[Json]:
```
aquamatthias, line 54: I would consider this a private method of the real azure client - not a public API.

aquamatthias (dismissed):

aquamatthias (approved):

PR #2051, opened 2024-04-30
[aws][feat]: Update `collect` method implementation

1101-1:
# Description
The `collect` method has been updated to make it easier to collect usage metrics.

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/fix_plugin_aws/resource/base.py
+++ b/plugins/aws/fix_plugin_aws/resource/base.py
@@ -225,6 +225,7 @@ class AwsResource(BaseResource, ABC):
         # - bend the json
         # - transform the result into a resource
         # - add the resource to the graph
+        # - return a list of resources
         # In case additional work needs to be done, override this method.
         instances = []
         for js in json:
@@ -233,7 +234,6 @@ class AwsResource(BaseResource, ABC):
                 instance.post_process(builder, js)
                 builder.add_node(instance, js)
                 instances.append(instance)
-
         return instances
 
     @classmethod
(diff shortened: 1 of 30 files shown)


plugins/aws/fix_plugin_aws/resource/iam.py
```diff
@@ -747,6 +749,7 @@ def add_virtual_mfa_devices() -> None:
 
         if builder.account.mfa_devices is not None and builder.account.mfa_devices > 0:
             builder.submit_work(service_name, add_virtual_mfa_devices)
+        return list(iam_users)
```
aquamatthias, line 752: No. This method collects not only users but groups roles etc.

plugins/aws/fix_plugin_aws/resource/base.py
```diff
@@ -220,20 +220,26 @@ def collect_resources(cls: Type[AwsResource], builder: GraphBuilder) -> None:
                 raise
 
     @classmethod
-    def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> None:
+    def collect(cls: Type[AwsResource], json: List[Json], builder: GraphBuilder) -> List[AwsResource]:
```
aquamatthias, line 223: I was on the fence to propose the following:
```python
    def collect(cls: Type[AwsResourceType], json: List[Json], builder: GraphBuilder) -> List[AwsResourceType]:
   buckets = AwsS3Bucket.collect()
```
buckets would have type AwsS3Bucket automatically.
But it would also mean, that each collect method will only collect instances of the related class.
But there are collect methods that also collect other resources.
See IAM user.

As a consequence: when instances of other kinds are collected, they should be returned.

aquamatthias (commented):
Nice. I

PR #2071, opened 2024-05-13
[aws][fix] Fix parallel collection

1101-1:
# Description

# To-Dos

# Code of Conduct


plugins/aws/fix_plugin_aws/collector.py
```diff
@@ -276,6 +288,25 @@ def work_done(_: Future[None]) -> None:
         when_done.add_done_callback(work_done)
         return when_done
 
+    def collect_usage_metrics(self, resources: List[Type[AwsResource]], builder: GraphBuilder) -> None:
+        futures: List[Future[Tuple[List[cloudwatch.AwsCloudwatchQuery], Dict[str, AwsResource], Dict[str, Any]]]] = [
+            builder.submit_work("cloudwatch", resource.collect_usage_metrics, builder) for resource in resources
+        ]
+        usage_metric_data = [future.result() for future in as_completed(futures)]
+        for data in usage_metric_data:
```
aquamatthias, line 296: No need to execute this in a future - only queries are created without performing any IO.
Instead of creating a list all queries, consider iterating the list until the next chunk is full (500 elements) and then you could spawn a future here.

plugins/aws/fix_plugin_aws/resource/base.py
```diff
@@ -220,28 +213,23 @@ def collect_resources(cls: Type[AwsResource], builder: GraphBuilder) -> None:
...
                 instance.post_process(builder, js)
                 builder.add_node(instance, js)
-                instances.append(instance)
-        return instances
 
     @classmethod
-    def collect_usage_metrics(
-        cls: Type[AwsResource], builder: GraphBuilder, collected_resources: List[AwsResource]
-    ) -> None:
+    def collect_usage_metrics(cls: Type[AwsResource], builder: GraphBuilder) -> Tuple[List, Dict[str, AwsResource], Dict[str, Any]]:  # type: ignore
```
aquamatthias, line 230: This signature is .. complicated and makes me think:
- it is a class method
- returns the list of objects to match
- returns queries and normalizers

Wouldn't it be much simpler, if the method were an instance method, not a class method? This way we would not need the first return value. If you would make the normalization part of the query, you then would only return this:
```
List[AwsCloudwatchQuery]
```

1101-1:
```python
def update_resource_metrics(
    resources_map: Dict[str, V@update_resource_metrics],
    cloudwatch_result: Dict[AwsCloudwatchQuery, AwsCloudwatchMetricData],
    metric_normalizers: Dict[str, MetricNormalization]
) -> None
```
`update_resource_metrics` requires  `resources_map`{class.id: class}, `cloudwatch_result`{query: result_data_query} and `metric_normalizers`{metric_name: metric_normalizator}

If we pass `metric_normalizers` in `AwsCloudwatchQuery`, then we can't use `update_resource_metrics`. Because we can't pass normilizer to the function in this case

aquamatthias (dismissed):

PR #2087, opened 2024-05-27
[fixlib][feat] Export categories

lloesche:
# Description

# To-Dos

# Code of Conduct

### DIFF
--- a/fixlib/fixlib/basecategories.py
+++ b/fixlib/fixlib/basecategories.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
+from typing import Any
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


meln1k (dismissed):

fixlib/fixlib/baseresources.py
```diff
@@ -719,6 +722,23 @@ def __getstate__(self) -> Dict[str, Any]:
...
+                    merged.extend(gather_categories(base))
+            if hasattr(class_type, "_categories"):
+                merged.extend(class_type._categories)
+            return merged
+
+        return list(set(gather_categories(cls)))
+
+    @property
+    def categories(self) -> List[str]:
+        return [str(category.value) for category in self.get_all_categories()]
```
aquamatthias, line 740:
```suggestion
        return [category.value.name for category in self.get_all_categories()]
```
To not rely on the str representation?

fixlib/fixlib/basecategories.py
```diff
@@ -0,0 +1,113 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any
+
+
+@dataclass(frozen=True)
+class BaseCategory:
```
aquamatthias, line 7: eq should be implemented by dataclass automatically.
You could also add @functools.total_ordering to the class and only implement __lt__.

aquamatthias (approved):
Two small remarks - you can ignore them if you don't like it.

fixlib/fixlib/baseresources.py
lloesche: I purposefully chose str() here (and anywhere else) because it allows me to change the (already existing) implementation of __str__ in one place and it changes throughout the code. If I use .name here and later decide I'd rather use .long_name or .description or something else as the string representation, I might miss it in one place and wouldn't necessarily see an error, just inconsistent data.

fixlib/fixlib/basecategories.py
lloesche: > __eq__ should be implemented by dataclass automatically.

It is, but it's not based on comparing str(self).

PR #2119, opened 2024-06-18
[aws][feat] Add tag deletion/updating and resource deletion for backup service

1101-1:
# Description

# To-Dos

# Code of Conduct


plugins/aws/fix_plugin_aws/resource/backup.py
```diff
@@ -121,7 +123,7 @@ def add_tags(backup_job: AwsBackupJob) -> None:
                 "list-tags",
                 "Tags",
                 expected_errors=["ResourceNotFoundException"],
-                ResourceArn=backup_job.arn,
+                ResourceArn=backup_job.recovery_point_arn,
```
aquamatthias, line 126: You would query for tags of the recovery point - that's wrong isn't it?
Does a job even have tags?

plugins/aws/fix_plugin_aws/resource/backup.py
```diff
@@ -558,6 +656,43 @@ def called_collect_apis(cls) -> List[AwsApiSpec]:
...
+
+    def delete_resource_tag(self, client: AwsClient, key: str) -> bool:
+        client.call(
+            aws_service=self.api_spec.service,
+            action="untag-resource",
+            result_name=None,
+            ResourceArn=self.arn,
+            TagKeyList=[key],
+        )
+        return True
```
aquamatthias, line 685: can we extract a separate base class? Seems to be repeated.
See CloudwatchTaggable as example

plugins/aws/fix_plugin_aws/resource/backup.py
```diff
@@ -771,7 +972,7 @@ def add_tags(restore_job: AwsBackupRestoreJob) -> None:
                 "list-tags",
                 "Tags",
                 expected_errors=["ResourceNotFoundException"],
-                ResourceArn=restore_job.arn,
+                ResourceArn=restore_job.recovery_point_arn,
```
aquamatthias, line 975: Same here.
You would query for tags of the recovery point - that's wrong isn't it?
Does a job even have tags?

plugins/aws/fix_plugin_aws/resource/backup.py
```diff
@@ -868,7 +1069,7 @@ def add_tags(copy_job: AwsBackupCopyJob) -> None:
                 "list-tags",
                 "Tags",
                 expected_errors=["ResourceNotFoundException"],
-                ResourceArn=copy_job.arn,
+                ResourceArn=copy_job.source_recovery_point_arn,
```
aquamatthias, line 1072: same here.

plugins/aws/fix_plugin_aws/resource/backup.py
1101-1: Tags from the original resource are automatically copied to the recovery points created during backup jobs. This ensures that the backup data retains the tagging information of the original resources.

https://docs.aws.amazon.com/aws-backup/latest/devguide/tags-on-backups.html

But yes, it is not a backup job/copy job/restore job tags

aquamatthias (approved):

PR #2150, opened 2024-07-23
[azure][feat] Add support for mysql resource collection

1101-1:
# Description
Add a MySQL resource collection implementation.

# To-Dos

# Code of Conduct


plugins/azure/fix_plugin_azure/resource/mysql.py
```diff
@@ -0,0 +1,771 @@
...
+            self.volume_encrypted = False
+
+    def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
+        if location := self.location:
+            builder.add_edge(
+                self, edge_type=EdgeType.default, clazz=AzureMysqlCapability, display_location_name=location
+            )
+            builder.add_edge(
+                self, edge_type=EdgeType.default, clazz=AzureMysqlCapabilitySet, display_location_name=location
+            )
```
aquamatthias, line 706: This would create edges to all capabilities and capability sets in the same region and is rather useless.
Ideally every MySQL instance would have exactly ONE "type" which describes ALL the capabilities of this server.
I would suggest creating such a resource if it does not exist.

aquamatthias (commented):

fixlib/fixlib/baseresources.py
```diff
@@ -1094,6 +1094,23 @@ class DatabaseInstanceStatus(Enum):
     UNKNOWN = "unknown"
 
 
+@define(eq=False, slots=False)
+class BaseDatabaseInstanceType(BaseType):
```
aquamatthias, line 1098: As it is now, can we just use BaseInstanceType?

aquamatthias (commented):
Nice. Only one question left.

fixlib/fixlib/baseresources.py
1101-1: We can do something like:
```python
class BaseDatabaseInstanceType(BaseInstanceType):
    kind: ClassVar[str] = "database_instance_type"
    kind_display: ClassVar[str] = "Database Instance Type"
    kind_description: ClassVar[str] = "A database instance type."
    metadata: ClassVar[Dict[str, Any]] = {"icon": "database", "group": "database"}
```
Here we change the metadata and `kind` properties, and other fields inherit from `BaseInstanceType`

PR #2155, opened 2024-07-29
[azure][feat] Add postgres service collection

1101-1:
# Description
Add a postgresql service collection

# To-Dos

# Code of Conduct


plugins/azure/fix_plugin_azure/resource/base.py
```diff
@@ -473,6 +473,102 @@ class AzureUserAssignedIdentity:
...
+@define(eq=False, slots=False)
+class AzureServerBackup:
+    kind: ClassVar[str] = "azure_server_backup"
+    mapping: ClassVar[Dict[str, Bender]] = {
+        "backup_interval_hours": S("backupIntervalHours"),
+        "backup_retention_days": S("backupRetentionDays"),
+        "earliest_restore_date": S("earliestRestoreDate"),
+        "geo_redundant_backup": S("geoRedundantBackup"),
+    }
+    backup_interval_hours: Optional[int] = field(default=None, metadata={'description': 'Backup interval hours for the server.'})  # fmt: skip
```
aquamatthias, line 510: why is this in base?

plugins/azure/fix_plugin_azure/resource/base.py
```diff
@@ -473,6 +473,102 @@ class AzureUserAssignedIdentity:
...
+
+@define(eq=False, slots=False)
+class AzureServerMaintenanceWindow:
+    kind: ClassVar[str] = "azure_server_maintenance_window"
+    mapping: ClassVar[Dict[str, Bender]] = {
+        "custom_window": S("customWindow"),
+        "day_of_week": S("dayOfWeek"),
+        "start_hour": S("startHour"),
+        "start_minute": S("startMinute"),
+    }
```
aquamatthias, line 550: base?

plugins/azure/fix_plugin_azure/resource/postgresql.py
```diff
@@ -0,0 +1,641 @@
...
+                                    "memoryPerVCoreMb": sku.supported_memory_per_vcore_mb,
+                                },
+                                "storage": {
+                                    "iops": storage.supported_iops,
+                                    "storageSizeGb": (
+                                        storage.storage_size_mb // 1024 if storage.storage_size_mb else 0
+                                    ),
+                                },
+                                "location": self.location,
+                            }
```
aquamatthias, line 253: We create here an arbitrary format which then is parsed in AzurePostgresqlServerType. 
Q1: what is the relationship between capability and server type? Can we get rid of one?
Q2: wouldn't it be simpler to create the server type based on the already parsed capability?

plugins/azure/fix_plugin_azure/resource/postgresql.py
```diff
@@ -0,0 +1,641 @@
...
+
+            AzurePostgresqlServerType.collect(server_types, graph_builder)
+
+        if self.supported_psql_flexible_server_editions:
+            for edition in self.supported_psql_flexible_server_editions:
+                graph_builder.submit_work(service_name, collect_editions, edition)
+
+
+@define(eq=False, slots=False)
+class AzurePostgresqlServerConfiguration(MicrosoftResource, AzureProxyResource):
```
aquamatthias, line 264: Same comment as on MySQL:
I would imagine something like this:

```python
class AzureDatabseConfiguration(MicrosoftResource):
   config: Optional[Json] = None
```

We would  read all ServerConfiguration Resources and turn them into a simple dict.

aquamatthias (changes requested):
Thanks. Please find my comments.

plugins/azure/fix_plugin_azure/resource/base.py
1101-1: We use `AzureServerBackup`, `AzureServerDataEncryption`, `AzureServerMaintenanceWindow`, `AzureServerHighAvailability` and `AzureServerNetwork` classes in `mysql` and `postgresql` resources.

plugins/azure/fix_plugin_azure/resource/postgresql.py
1101-1: We collect capabilities. Then from capabilities we create a `AzurePostgresqlServerType`/`AzureMysqlServerType`. After all works done, we remove capabilities from the graph.

aquamatthias: Consider removing the capability class and read the related type directly.
This way we do not have an artificial type in our model, nor do we need to maintain 2 mappings.

plugins/azure/fix_plugin_azure/resource/base.py
aquamatthias: I would try to keep the base as minimal as possible. Importing classes from MySQL is most probably fine in PostgreSQL (or vice versa) .

aquamatthias (approved):

PR #2211, opened 2024-09-25
[gcp][feat] Add Vertex AI collection

1101-1:
# Description

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/gcp/fix_plugin_gcp/resources/aiplatform.py
+++ b/plugins/gcp/fix_plugin_gcp/resources/aiplatform.py
@@ -872,7 +872,9 @@ class GcpVertexAIDatasetVersion(AIPlatformRegionFilter, GcpResource):
     )
     kind_service: ClassVar[Optional[str]] = service_name
     metadata: ClassVar[Dict[str, Any]] = {"icon": "resource", "group": "ai"}
-    reference_kinds: ClassVar[ModelReference] = {"predecessors": {"default": ["gcp_vertex_ai_model"]}}
+    reference_kinds: ClassVar[ModelReference] = {
+        "predecessors": {"default": ["gcp_vertex_ai_model", GcpVertexAIDataset.kind]}
+    }
     api_spec: ClassVar[GcpApiSpec] = GcpApiSpec(
         service="aiplatform",
         version="v1",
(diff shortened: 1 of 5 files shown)


plugins/gcp/fix_plugin_gcp/resources/aiplatform.py
```diff
@@ -0,0 +1,3376 @@
...
+    def connect_in_graph(self, builder: GraphBuilder, source: Json) -> None:
+        if tuned_model := self.tuned_model:
+            if endpoint := tuned_model.endpoint:
+                builder.add_edge(self, reverse=True, clazz=GcpVertexAIEndpoint, name=endpoint)
+
+
+resources: List[Type[GcpResource]] = [
+    GcpVertexAIModel,
+    GcpVertexAIDataset,
+    # GcpVertexAIDatasetVersion, : collected via GcpVertexAIFeatureGroup
```
aquamatthias, line 3359: can we uncomment them? So we can get the list of resources. 
It is fine if the resources are collected via another resource.

plugins/gcp/fix_plugin_gcp/resources/aiplatform.py
```diff
@@ -0,0 +1,3376 @@
...
+    "southamerica-west1",
+    "us-central1",
+    "us-east1",
+    "us-east4",
+    "us-south1",
+    "us-west1",
+    "us-west2",
+    "us-west3",
+    "us-west4",
+]
```
aquamatthias, line 60: There is no good way to synchronize this information. I would make it more stupid: let's collect in all regions and ignore the errors if not existent.

plugins/gcp/fix_plugin_gcp/resources/aiplatform.py
1101-1: We use method `collect_resources` on each of commented resources. for example:
```python
class GcpVertexAIDataset(AIPlatformRegionFilter, GcpResource):
    @classmethod
    def collect(cls: Type[GcpResource], raw: List[Json], builder: GraphBuilder) -> List[GcpResource]:
        # Additional behavior: iterate over list of collected GcpVertexAIDataset and for each:
        # - collect related GcpVertexAIDatasetVersion
        result: List[GcpResource] = super().collect(raw, builder)  # type: ignore
        dataset_ids = [dataset.id for dataset in cast(List[GcpVertexAIDataset], result)]
        for dataset_id in dataset_ids:
            builder.submit_work(GcpVertexAIDatasetVersion.collect_resources, builder, parent=dataset_id)

        return result
```
If we uncomment these sub-resources, then collector will use their `api_spec` and will not be able to collect the data.

aquamatthias: Mhh I am sure that would be solvable - it is not that important to block this PR.

aquamatthias (approved):

PR #2258, opened 2024-10-17
Improve access edge collection performance

meln1k:
# Description

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/aws/fix_plugin_aws/access_edges.py
+++ b/plugins/aws/fix_plugin_aws/access_edges.py
@@ -2,9 +2,9 @@ from functools import lru_cache
 from attr import frozen, define
 import networkx
 from fix_plugin_aws.resource.base import AwsAccount, AwsResource, GraphBuilder
-
+from policy_sentry.querying.actions import get_actions_for_service
 from typing import Dict, List, Literal, Set, Optional, Tuple, Union, Pattern
-
+import fnmatch
 from networkx.algorithms.dag import is_directed_acyclic_graph
 
 from fixlib.baseresources import (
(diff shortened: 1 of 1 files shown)


plugins/aws/fix_plugin_aws/access_edges.py
```diff
@@ -102,17 +102,60 @@ def find_all_allowed_actions(all_involved_policies: List[PolicyDocument], resour
...
+        return []
 
 
 @lru_cache(maxsize=1024)
+def get_expanded_action(statement: StatementDetail, service_prefix: str) -> List[str]:
+    expanded: List[str] = statement_expanded_actions(statement, service_prefix)
+    return expanded
+
+
+@lru_cache(maxsize=10000)
```
aquamatthias, line 158: Instead of creating rather stupid regexes from the filters, consider parsing it into an ARN Filter:
`arn:partition:service:region:account-id:resource-type/resource-id`
Every part would be one of:
- fixed  (e.g. `foo`)
-match_all (e.g. `*`)
-pattern (e.g. `foo*bla`)

Applying this filter to the resource is potentially much faster (especially with `*`) and only needs a regexp search in case of pattern.

plugins/aws/fix_plugin_aws/collector.py
```diff
@@ -271,7 +262,7 @@ def get_last_run() -> Optional[datetime]:
                     log.warning(f"Unexpected node type {node} in graph")
                     raise Exception("Only AWS resources expected")
 
-            access_edge_collection_enabled = os.environ.get("ACCESS_EDGE_COLLECTION_ENABLED", "false").lower() == "true"
+            access_edge_collection_enabled = False
```
aquamatthias, line 265: This basically disables the functionality always. Let's use it this way: 
- Remove this flag
- config defaults to false
- The config property has a warning as a description.

If someone wants to turn it on, it must be enabled in the configuration with clear documentation.

aquamatthias (changes requested):

PR #2284, opened 2024-11-29
[feat][gcp] Improve the way of collection for DiskTypes and MachineTypes

1101-1:
# Description

# To-Dos

# Code of Conduct

### DIFF
--- a/plugins/gcp/fix_plugin_gcp/resources/base.py
+++ b/plugins/gcp/fix_plugin_gcp/resources/base.py
@@ -217,13 +217,13 @@ class GraphBuilder:
                     node._zone = zone
                     node._region = self.region_by_zone_name.get(zone.id)
                     self.add_edge(zone, node=node)
-                    return True
+                    return
 
                 # Then check for region
                 if region := self.region_by_name.get(location_name):
                     node._region = region
                     self.add_edge(region, node=node)
-                    return True
+                    return
 
         if source is not None:
             if ZoneProp in source:
@@ -234,7 +234,7 @@ class GraphBuilder:
                     self.add_edge(node, node=zone, reverse=True)
                     return
                 else:
-                    log.warning(
+                    log.debug(
                         "Zone property '%s' found in the source but no corresponding zone object is available to associate with the node.",
                         zone_name,
                     )
@@ -246,7 +246,7 @@ class GraphBuilder:
                     self.add_edge(node, node=zone, reverse=True)
                     return
                 else:
-                    log.warning(
+                    log.debug(
                         "Internal zone property '%s' exists in the source but no corresponding zone object is available to associate with the node.",
                         source[InternalZoneProp],
                     )
@@ -258,7 +258,7 @@ class GraphBuilder:
                     self.add_edge(node, node=region, reverse=True)
                     return
                 else:
-                    log.warning(
+                    log.debug(
                         "Region property '%s' found in the source but no corresponding region object is available to associate with the node.",
                         region_name,
                     )
--- a/plugins/gcp/fix_plugin_gcp/resources/base.py
+++ b/plugins/gcp/fix_plugin_gcp/resources/base.py
@@ -187,27 +187,26 @@ class GraphBuilder:
                     result.append(n)
         return result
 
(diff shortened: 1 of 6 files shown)


plugins/gcp/fix_plugin_gcp/resources/base.py
```diff
@@ -173,25 +173,33 @@ def _standard_edges(self, node: GcpResourceType, source: Optional[Json] = None)
...
 
         if source is not None:
+            if "zone" in source:
+                zone_name = source["zone"].rsplit("/", 1)[-1]
+                if zone := self.zone_by_name.get(zone_name):
+                    node._zone = zone
+                    node._region = self.region_by_zone_name[zone_name]
+                    self.add_edge(node, node=zone, reverse=True)
+                    return True
+            if "region" in source:
```
aquamatthias, line 183: Isn't that case already handled on line 197?

plugins/gcp/fix_plugin_gcp/resources/base.py
```diff
@@ -173,25 +173,33 @@ def _standard_edges(self, node: GcpResourceType, source: Optional[Json] = None)
             return True
 
         if source is not None:
+            if "zone" in source:
```
aquamatthias, line 176:
```suggestion
            if ZoneProp in source:
```
to align with the other props

plugins/gcp/fix_plugin_gcp/resources/base.py
```diff
@@ -173,25 +173,33 @@ def _standard_edges(self, node: GcpResourceType, source: Optional[Json] = None)
...
+                    self.add_edge(node, node=region, reverse=True)
+                    return True
+
             if InternalZoneProp in source:
                 if zone := self.zone_by_name.get(source[InternalZoneProp]):
                     node._zone = zone
                     node._region = self.region_by_zone_name[source[InternalZoneProp]]
                     self.add_edge(node, node=zone, reverse=True)
                     return True
-                else:
```
aquamatthias, line 182: why remove? a zone prop was defined, but no zone is found.

plugins/gcp/fix_plugin_gcp/resources/base.py
```diff
@@ -173,25 +173,33 @@ def _standard_edges(self, node: GcpResourceType, source: Optional[Json] = None)
...
-                    log.debug(f"Zone {source[InternalZoneProp]} not found for node: {node}. Ignore resource.")
-                    return False
 
             if RegionProp in source:
                 region_name = source[RegionProp].rsplit("/", 1)[-1]
                 if region := self.region_by_name.get(region_name):
                     node._region = region
                     self.add_edge(node, node=region, reverse=True)
                     return True
-                else:
```
aquamatthias, line 192: why remove? a region prop was defined, but no zone is found.

aquamatthias (changes requested):

plugins/gcp/fix_plugin_gcp/resources/base.py
1101-1: If we return False, then we will not add the node to the graph:
```python
    def add_node(self, node: GcpResourceType, source: Optional[Json] = None) -> Optional[GcpResourceType]:
       ...
        if self._standard_edges(node, source):
            with self.graph_nodes_access:
                self.graph.add_node(node, source=source or {})
            return node
        return None
```
Also we have the another checking if the graph_builder has region property:
```python
        # Fallback to GraphBuilder region, i.e. regional collection
        if self.region is not None:
            node._region = self.region
            self.add_edge(node, node=self.region, reverse=True)
            return True
 ```
In case if region still not found, we just add the `global` region:
```python
        # Fallback to global region
        node._region = self.fallback_global_region
        self.add_edge(node, node=self.fallback_global_region, reverse=True)
        return True
```
This way the node will always be added to the graph

aquamatthias: If a zone prop is defined but the related zone is not found, the builder will attach it to its region.
The error will silently be dropped. 
We should at least log a warning in such a case.

aquamatthias (approved):

PR #2321, opened 2025-03-28
Fix some AGPL references that were left after returning to Apache2 

monoman:
Just to make the intention clear although License classifiers are deprecated...


lloesche (approved):
Thank you @monoman for the PR. Good catch!

monoman:
😃

